PR #5, opened 2017-03-06
was failing to compile linux - error: unknown type name ‘uint16_t’

scottstensland:
just added header and a linker math library to fix make compile errors

after these changes was able to successfully compile on 

64 bit  Ubuntu 16.04 
gcc version 5.4.1 20160904 (Ubuntu 5.4.1-2ubuntu1~16.04)


marcobambini:
Thanks a lot!

PR #13, opened 2017-03-07
Some Grammar Fixes in the Docs

hallzy:
Just a few grammatical fixes I made in the docs.

As for the "indices" vs "indexes", while both words are correct plural forms of index, it is probably best to keep it consistent, so I just changed two instances of "indexes" to "indices"


marcobambini:
Thanks a lot I really appreciate!

PR #14, opened 2017-03-07
Added "gravity" and "unittest" executables to gitignore

hallzy:
These generated executables probably should be in the gitignore.

Up to you though.


marcobambini:
You are right, thanks!

PR #22, opened 2017-03-07
_int64 does not exist but __int64 on WIN32

GeertVL:
Otherwise it does not compile on WIN32.

### DIFF
--- a/src/utils/gravity_utils.h
+++ b/src/utils/gravity_utils.h
@@ -13,7 +13,7 @@
 #include <stdbool.h>
 
 #if defined(_WIN32)
-typedef unsigned _int64 nanotime_t;
+typedef unsigned __int64 nanotime_t;
 #define DIRREF			HANDLE
 #else
 #include <dirent.h>


marcobambini:
Thanks a lot, Windows version really needs some more attention!

GeertVL:
Marco, 



No problem. I am working on the Windows version. Great language. A lot of potential. 
I am learning a lot about languages in the mean time. 

I will work now file per file to get the Windows part working. 



Geert 


Van: "Marco Bambini" <notifications@github.com> 
Aan: "marcobambini/gravity" <gravity@noreply.github.com> 
Cc: "Geert Van Laethem" <geert.van.laethem@pandora.be>, "Author" <author@noreply.github.com> 
Verzonden: Dinsdag 7 maart 2017 13:50:13 
Onderwerp: Re: [marcobambini/gravity] _int64 does not exist but __int64 on WIN32 (#22) 



Thanks a lot, Windows version really needs some more attention!

GeertVL:
Are we working with Wide or Ansi for the Windows part? 
For Windows Wide is the recommended string format.

Is that ok that I change everything to wide as it is inconsistent now?

marcobambini:
I would prefer to stay Ansi because internally the lexer is already able to recognize and properly handle UTF-8 characters.

GeertVL:
@marcobambini 

Ok I will go to ANSI. No problem.

Issue #26, opened 2017-03-07
Visual Studio Code syntax


Dohxis:
Is anyone interested in extension for Visual Studio Code for syntax highlighting? I have simple extension already working, with minor modifications I could add it on Github if anyone is interested.

marcobambini:
I can add a new folder to the official repository to collect syntax highlighting definitions.
Thanks!

Dohxis:
Just published an extension, you can find it on vscode [marketplace](https://marketplace.visualstudio.com/items?itemName=Dohxis.gravity)! Only simple syntax highlighter for now.

marcobambini:
Very well done, thanks!

Issue #23, opened 2017-03-07
ide or editor support


wangtxxl:
this project is very good  china people need it  ！！There is a market in China  ，keep it light optimization speed

darthdeus:
Having a syntax file for VIM would definitely be a good start :)

hallzy:
For anyone looking for vim syntax highlighting:

Note that the PR referenced above was reverted. The URL for the vim gravity syntax can be found in the [wiki](https://github.com/marcobambini/gravity/wiki). So far it is essentially a swift syntax highlighter, but I welcome changes and additions.

PR #30, opened 2017-03-07
Added vim syntax highlighting.

hallzy:
Reference #26 and #23 for syntax highlighting.

This is copied from

* https://github.com/keith/swift.vim

The original author requires that copyright notice to be in his work
when used elsewhere.

Also, this is unmodified (except for substitutions of "swift" with "gravity",
so it is actually a highlighting for swift.

It works well for gravity currently though, and we can change it as
needed.

### DIFF
--- a/syntax_highlighting/vim/ftdetect/gravity.vim
+++ b/syntax_highlighting/vim/ftdetect/gravity.vim
@@ -1,0 +1,32 @@
+" Copyright (c) 2014 Keith Smiley (http://keith.so)
+" Permission is hereby granted, free of charge, to any person obtaining a copy
+" of this software and associated documentation files (the 'Software'), to deal
+" in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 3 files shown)


hallzy:
Unless, you would now rather not have syntax highlighting in this repo? In which case, we can setup a separate repo for this vim highlight

marcobambini:
Thanks a lot!

marcobambini:
Hi @hallzy now that I started an official Wiki https://github.com/marcobambini/gravity/wiki do you mind to setup a separate repo for the vim highlighter?

hallzy:
yes of course. You want to revert this commit then?

The repo with the vim syntax is: https://github.com/hallzy/gravity.vim

marcobambini:
Yes @hallzy I am going to revert this commit and add your repo in the [wiki](https://github.com/marcobambini/gravity/wiki).

marcobambini:
Thanks a lot!

PR #34, opened 2017-03-07
Windows build - Fix file read/write and windows includes

bendl:
Building with:
`mingw32-make CC=gcc`

This will successfully build both `gravity.exe` and `unittest.exe`.
I have tested gravity.exe with examples on https://marcobambini.github.io/gravity/syntax.html and it works OK. Linux build is unbroken (tested on Windows 10 bash).

This has uncovered a bug with the `unittest` project where there is a problem with `directory_read` and `FindNextFile` on win32 and is not finding files, but thats for another issue/PR.


marcobambini:
Very well done, thanks @bendl!

PR #39, opened 2017-03-07
Remove warnings

hallzy:
This should remove all the warnings that get produced from make.

All tests pass for me with this change.

I forgot to disable the part of my text editor that removes trailing whitespace, so there are a lot of those changes. If you would prefer to not have the whitespace removed in these commits, let me know and I will go back through and just make the actual changes (since the whitespace changes account for a lot of modifications)

The `-fgnu89-inline` flag in the Makefile tells GCC to use the traditional GNU semantics for inline functions when in C99 mode, thus eliminating the warnings produced by the inline functions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ SRC = $(wildcard $(COMPILER_DIR)*.c) \
       $(wildcard $(UTILS_DIR)/*.c)
 
 INCLUDE = -I$(COMPILER_DIR) -I$(RUNTIME_DIR) -I$(SHARED_DIR) -I$(UTILS_DIR)
-CFLAGS = $(INCLUDE) -O2 -std=gnu99
+CFLAGS = $(INCLUDE) -O2 -std=gnu99 -fgnu89-inline
 OBJ = $(SRC:.c=.o)
 
 ifeq ($(OS),Windows_NT)
--- a/src/compiler/gravity_ast.c
+++ b/src/compiler/gravity_ast.c
@@ -6,6 +6,7 @@
 //  Copyright (c) 2014 CreoLabs. All rights reserved.
 //
 
+#include <inttypes.h>
 #include "gravity_ast.h"
 #include "gravity_utils.h"
 #include "gravity_visitor.h"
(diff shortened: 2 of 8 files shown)


marcobambini:
Well done @hallzy but yes please remove the trailing whitespace changes because too many files seems changed without any real modification.

hallzy:
That should do it now. All whitespace changes removed, still passes all the tests.

marcobambini:
@hallzy sorry to be a bit meticulous but for the 2 changes in codegen.c I think the it would be better to fix with:
uint16_t nparams = (node->params) ? (uint16_t)marray_size(*node->params) : 0;
instead of
uint16_t nparams = (uint16_t*)(node->params) ? marray_size(*node->params) : 0;

and about the %lld changed with %ld are you really sure about it?
Just asking because I thought that %lld should be used for 64 bit int but I could be wrong here.

hallzy:
You have every right to be meticulous :)


This seems to work:

```c
uint16_t nparams = (node->params) ? (uint16_t)marray_size(*node->params) : 0;
```

The compiler doesn't like it when you do `node->params` instead of `*node->params` though (results in an error), so it seems we do still need the "*"

As for the `%lld` vs `%ld`, I am not sure what it shows when you try to compile it, but for me the warning was:

```
warning: format ‘%lld’ expects argument of type ‘long long int’, but argument 4 has type ‘long int’ [-Wformat=]
```

So in my warning at least, it seems to suggest that %lld is for long long int, while %ld is for long int.

It is funny though, because looking into it a bit more, I see people supporting either of them. From what I have just found, it seems that "long long int" is at least 64 bit, but could be higher.

I have seen some people say that instead of `"%ld"` or `"%lld"`, you can also use `"%" PRId64` for int64_t, and `"%" PRIu64` for uint64_t

Using this requires `#include <inttypes.h>` though.

I guess it is up to you how you would like to handle this. May I ask, do you get warnings for the lld when you compile?

hallzy:
it actually looks like long long int and long int are both 64 bit on my computer... at least, that is what it seems, when looking at a sizeof

hallzy:
I updated this PR with that PRId64 thing so you can see how it works. Still compiles, all tests still pass

marcobambini:
Seems like the right solution, even if I wondering if Windows supports that notation.
Anyway, thanks @hallzy!

PR #58, opened 2017-03-10
Visual Studio solution file and supporting stuff

cartman300:
It builds ``bin/GravityLang.dll`` with exported symbols which are defined in ``gravity_visualstudio/exports.def``

It builds with v141 (VS 17) and v140 (VS 2015) toolsets which i tried.


marcobambini:
Great work @cartman300 but it seems that current .gitignore file content is complete deleted and replaced by Windows only settings. Can you confirm?

cartman300:
It includes the original entries after this line

https://github.com/cartman300/gravity/blob/eea6ca64c6d51ab42056cc8651f2345e8162977b/.gitignore#L254

marcobambini:
Thanks a lot @cartman300 Visual Studio integration is a very important milestone.
Please don't forget to add your name to the official CONTRIBUTORS file.

PR #65, opened 2017-03-11
Fix indentation in syntax.html

hkdnet:
It seems there are some indentation errors at "Gravity: Syntax" > "Blocks and Scope"
see: https://marcobambini.github.io/gravity/syntax.html
1 tab is used in other code blocks.


marcobambini:
Thanks @hkdnet!

PR #77, opened 2017-03-14
Added C++ include support and created Makefile command to build a shared object.

brandon-ray:
Added ```extern "C" {``` to fix include issues with C++ builds. Also added ```make lib``` to build a shared object that can be linked with C++ programs.


marcobambini:
Well done @brandon-ray, thanks.

Issue #62, opened 2017-03-10
Module system suggestions


cartman300:
Besides some way to load other modules written in gravity, i would suggest implementing the functionality to load native libraries as modules too.

For example ``module "test"`` would load ``test.gravity`` but in case that does not exist it would try ``test.dll`` or ``test.so`` and load a global ``gravity_load`` function where you can register functions/classes using the native API.

nanalan:
```ruby
import "test" # module named test, see other gh issues
import "test.gravity"
import "test.so"
import "test.dll"
import "./test" # looks for test.gravity, then test.dll, then test.so?
```

cartman300:
Yes, except i would make the file extensions optional and only resolve native libraries after nothing written in gravity has been found. That way the language can be extended with system dependent functionality on different systems and the common gravity codebase wouldn't have to change.

nanalan:
@cartman300 yep, that's what I was getting at with `import "./test"`.

GreenJoey:
Will exposing native C API be really good?? Should Gravity allow external code to randomly poke at a running interpreter??? Nobody developing a new modern language seems to be doing it!!!! Swify doesn't does it nor does Go neither does PyPy!!! I will basically make breaking changes in the interpreter very hard!!! Also exposing GC is probably not good as the native extension writer may fail to comply with GC standards!! Please don't expose native API

Mijyuoon:
@GreenJoey So you're basically saying that native modules shouldn't be supported because some developers are retards or what?

cartman300:
Literally every language you posted as example supports native code interop in one way or another.

GreenJoey:
Go doesn't let you call C directly but via a built-in C FFI. PyPy also does allow C code but via **C FFI**. And you can't assume that every programmer is perfectly normal. I am suggesting developers not to support native modules but support calling C functions via FFI.

marcobambini:
I like the idea to add support for FFI (in addition to the existing C API). In this way users will be able to call c code specifying its prototype from within Gravity.

parro-it:
> import "test" # module named test. See issue #37
> import "test.gravity"
> import "test.so"
> import "test.dll"
> import "./test" # looks for test.gravity, then test.dll, then test.so?

@nanalan we should also need a file extension for json bytecode, I think gravity should load also pre-compilend, non-native modules.

nanalan:
> we should also need a file extension for json bytecode

[...]

PR #80, opened 2017-03-14
Added documentation for the import statement.

hallzy:
Also, removed that task from the TODO list

I stuck it into the syntax file, but perhaps you want it somewhere else? Maybe its own import.html file or something?

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -7,6 +7,5 @@
 * Improve parser error handling
 
 **DOCUMENTATION AND EXAMPLES**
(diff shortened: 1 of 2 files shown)


marcobambini:
Thanks @hallzy I merged your pull request.
Import documentation was not added because I'd like to finalize the command first (see https://github.com/marcobambini/gravity/issues/62) but it is fine to have a preliminary doc in the syntax section.

hallzy:
Good to know, thanks

Issue #81, opened 2017-03-14
Support bitwise operators on boolean


leonardosnt:
This code
```swift
func main() {
  var boolean = true;
  boolean &= false;

  return boolean;
}
```
Causes a runtime error.
```
RUNTIME ERROR: Unable to perform operator BAND on object
```

hallzy:
I will have a pr for this soon.

marcobambini:
Fixed by https://github.com/marcobambini/gravity/commit/4cb3f502b7cd40cc1a2a9183cb5d02c18e0185e7 thanks to @hallzy

PR #82, opened 2017-03-15
Added support for bitwise operations on booleans

hallzy:
Ref: #81 

This excludes the `<<` and `>>` operations, since that would kind of go against what a bool is in my opinion.


marcobambini:
Thanks for the commit, bool case should also be added to the CHECK_FAST_BINARY_BIT macro in order to speed up these operations on bool.

hallzy:
Sorry, I am not sure I know what you mean. Do you mean modifying the macro on line 125 of `src/runtime/gravity_vmmacros.h` ?

It is quite late where I am so I probably won't get this done tonight.

marcobambini:
Already done with https://github.com/marcobambini/gravity/commit/aa8c24fc6ba958f5a39d92ed89c211d79372a51f

PR #90, opened 2017-03-16
Preemption [WIP] [RFC]

sargun:
This PR introduces basic preemption support into Gravity. I'd mostly like to get your thoughts on this. I'd also like to make it so that when (certain) native functions are called, it can also trigger preemption.

### DIFF
--- a/src/cli/gravity.c
+++ b/src/cli/gravity.c
@@ -131,6 +131,19 @@ static void gravity_repl (void) {
 
 // MARK: -
 
+int icounter = 0;
+
+static bool should_preempt(gravity_vm *vm, void *xdata) {
+	icounter++;
+	// Preempt ever 5 ins
+	if (icounter % 5 == 0) {
+		return true;
+	}
+	return false;
+}
+
+// MARK: -
+
 int main (int argc, const char* argv[]) {
 	// parse arguments and return operation type
 	op_type type = parse_args(argc, argv);
@@ -158,6 +171,7 @@ int main (int argc, const char* argv[]) {
 	
 	// create VM
 	gravity_vm *vm = gravity_vm_new(&delegate);
+	delegate.preempt_callback = &should_preempt;
 	
 	// check if input file is source code that needs to be compiled
 	if ((type == OP_COMPILE) || (type == OP_COMPILE_RUN)) {
@@ -198,8 +212,15 @@ int main (int argc, const char* argv[]) {
 	
 	// sanity check
 	assert(closure);
-	
-	if (gravity_vm_runmain(vm, closure)) {
+
+	bool result;
+	result = gravity_vm_runmain(vm, closure, true);
+
+	while (!result && gravity_vm_preempted(vm)) {
+		result = gravity_vm_run_resume_main(vm, closure, true);
+	}
+
+	if (result) {
 		gravity_value_t result = gravity_vm_result(vm);
 		double t = gravity_vm_time(vm);
 		
--- a/src/runtime/gravity_vm.c
+++ b/src/runtime/gravity_vm.c
@@ -6,6 +6,7 @@
 //  Copyright (c) 2014 CreoLabs. All rights reserved.
 //
 
+#include <gravity_delegate.h>
 #include "gravity_hash.h"
 #include "gravity_array.h"
 #include "gravity_debug.h"
@@ -26,6 +27,15 @@ static void gravity_gc_transform (gravity_hash_t *hashtable, gravity_value_t key
 // Internal cache to speed up common operations lookup
 static uint32_t cache_refcount = 0;
 static gravity_value_t cache[GRAVITY_VTABLE_SIZE];
+
+struct vm_state {
+	gravity_fiber_t				*fiber;
+	gravity_delegate_t			*delegate;
+	gravity_callframe_t			*frame;
+	gravity_function_t			*func;
+	gravity_value_t				*stackstart;
+	uint32_t			*ip;
+};
 
 // Opaque VM struct
 struct gravity_vm {
@@ -65,6 +75,9 @@ struct gravity_vm {
 	double				tstat[GRAVITY_LATEST_OPCODE];		// internal used to collect microbenchmarks
 	nanotime_t			t;									// internal timer
 	#endif
+
+	bool                preempted;                         // Set to true if the VM was preempted -- check flag if returns false
+	struct vm_state     vm_state;
 };
 
 // MARK: -
@@ -275,7 +288,7 @@ static void gravity_vm_loadclass (gravity_vm *vm, gravity_class_t *c) {
 
 // MARK: -
 
-static bool gravity_vm_exec (gravity_vm *vm) {
+static bool gravity_vm_exec(gravity_vm *vm, bool preemptable) {
 	DECLARE_DISPATCH_TABLE;
 	
 	gravity_fiber_t				*fiber = vm->fiber;			// current fiber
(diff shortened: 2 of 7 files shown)


marcobambini:
May I ask you why you'd like to introduce preemption?

sargun:
I tried to discuss in Gitter. It might be easier there.

sargun:
I'm introducing concurrency to Gravity, with the following semantics:

1. Every gravity program will belong to a processlet. 
2. Every processlet has a 64-bit integer ID that's derived from incrementing a number.
3. Processlets do not share memory.
4. You can communicate between two processlets by doing a `send(ProcessletID, Message)` OR a `receive(timeout)`
5. A processlet can monitor the life of another processlet by using a `monitor(ProcessletID)`

src/cli/gravity.c
```diff
@@ -158,6 +171,7 @@ int main (int argc, const char* argv[]) {
 	
 	// create VM
 	gravity_vm *vm = gravity_vm_new(&delegate);
+    delegate.preempt_callback = &should_preempt;
```
InusualZ: Formatting

src/runtime/gravity_vm.c
```diff
@@ -286,16 +299,36 @@ static bool gravity_vm_exec (gravity_vm *vm) {
...
-	// load current callframe
-	LOAD_FRAME();
-	DEBUG_CALL("Executing", func);
-	
-	// sanity check
-	if ((ip == NULL) || (!func->bytecode) || (func->ninsts == 0)) return true;
-	
-	DEBUG_STACK();
-	
+    gravity_preempt_callback    preempt_cb = NULL;          // Callback to cause preemption
```
InusualZ: ditto

src/runtime/gravity_vm.c
```diff
@@ -1407,7 +1460,7 @@ void gravity_vm_loadclosure (gravity_vm *vm, gravity_closure_t *closure) {
 	gravity_fiber_reassign(vm->fiber, closure, 0);
 	
 	// execute $moduleinit in order to initialize VM
-	gravity_vm_exec(vm);
+    gravity_vm_exec(vm, false);
```
InusualZ: ditto

src/runtime/gravity_vm.h
```diff
@@ -21,12 +21,15 @@ typedef void (*vm_transfer_cb) (gravity_vm *vm, gravity_object_t *obj);
 typedef void (*vm_cleanup_cb) (gravity_vm *vm);
 
 gravity_vm			*gravity_vm_new (gravity_delegate_t *delegate);
+void                gravity_vm_init (gravity_vm *vm, gravity_delegate_t *delegate);
```
InusualZ: ditto

marcobambini:
Have you measured the impact of performing MAYBE_PREEMPT on each instruction?

sargun:
It's negligible, performance wise to invoke a callback which returns false. It's about a 3% overhead versus commenting out MAYBE_PREEMPT entirely.

sargun:
The other place where need preemption / resumption support is native calls. I have yet to determine the interface here, but when making a call to a native function, it needs to be able to yield up, and resume. I haven't dug in very much, but I was thinking of introducing a return code from native functions, or another function that's bound to them that allows for yielding.

marcobambini:
Why a callback to check if VM must be interrupted and not a simpler:
`gravity_vm_interrupt`
that internally set a bool flag?

I would image APIs like:
```c
bool gravity_vm_interrupt(gravity_vm *vm, gravity_vm_state *state);
bool gravity_vm_resume(gravity_vm *vm, gravity_vm_state *state);
```

In this way no extra parameters are needed and the API is much more cleaner.

sargun:
I thought about such an API. It's not a bad approach, I thought about it initially -- especially invoking it from native C calls. I have some questions though:

1. What calls `vm_interrupt` / `vm_suspend`? If it's not being invoked through a callback, it needs to be a thread. Some of the platforms I intend to target do not readily have threads available -- highly sandboxed VMs, and WebASM / Emscripten.
2. When `gravity_vm_interrupt` gets called, does it unwind the stack of the VM? If so, how does the VM get resumed? If not, how does the VM yield?
3. If the boolean is being read locally, and modified by an external thread, we would have to declare it volatile, and I think that introducing synchronization would actually make it slower in some cases -- I can benchmark this if you want. 

----
As a side note, I tried out one other approach -- rather than a run-time callback, doing a compile-time macro. Unfortunately, this had no large speed difference as compared to the callback (<1% faster).

marcobambini:
Something like this should work in my opinion:

```c
bool gravity_vm_interrupt (gravity_vm *vm) {
    // if vm already interrupted return false
    if (vm->interrupted) return false;

    // check for vm_state callback
    if (!vm->delegate->vm_state) return false;

    // don't think we need to protect this bool flag
    vm->interrupted = true;

    // wait for vm_state callback to be called
    // automatically called by vm when flag is set
    return true;
}

void vm_state (gravity_vm *vm, gravity_vm_state *state) {
    // this callback is automatically called when the vm is interrupted
    // state contains info necessary to resume vm (like your vm_state struct)
}

bool gravity_vm_resume(gravity_vm *vm, gravity_vm_state *state) {
    // if vm not interrupted return false
    if (!vm->interrupted) return false;

    // reset interrupted flag 
    vm->interrupted = false;

    // resume execution
    gravity_vm_run_resume_main(...);    

    return true;
}

// your MAYBE_PREEMPT macro is now
#define MAYBE_PREEMPT     if (vm->interrupted) goto DO_PREEMPT;
```

marcobambini:
The ability to be able to interrupt the VM (and step) will open us the possibility to attach a debugger to it.

Jezza:
If you don't have threads, in which to interrupt the VM, why not look into a yield mechanic, or something similar?
Granted, that doesn't have the exact same functionality, unless you want to introduce a language level yield, but again, you would differ in functionality. 

Why not a "debug" opcode that halts the VM, also potentially reducing back pressure.

src/runtime/gravity_vm.c
```diff
@@ -1261,23 +1294,42 @@ static bool gravity_vm_exec (gravity_vm *vm) {
...
+	vm->vm_state.func = func;
+	vm->vm_state.ip = ip;
+	vm->vm_state.stackstart = stackstart;
+
+	return false;
 }
 
-gravity_vm *gravity_vm_new (gravity_delegate_t *delegate/*, uint32_t context_size, gravity_int_t gcminthreshold, gravity_int_t gcthreshold, gravity_float_t gcratio*/) {
+gravity_vm *gravity_vm_new(
+		gravity_delegate_t *delegate/*, uint32_t context_size, gravity_int_t gcminthreshold, gravity_int_t gcthreshold, gravity_float_t gcratio*/) {
```
InusualZ: Formatting

src/runtime/gravity_vm.c
```diff
@@ -1514,15 +1567,23 @@ bool gravity_vm_runmain (gravity_vm *vm, gravity_closure_t *closure) {
...
-	bool result = gravity_vm_exec(vm);
+	bool result = gravity_vm_exec(vm, preemptable);
 	nanotime_t tend = nanotime();
-	vm->time = millitime(tstart, tend);
-	
-	PRINT_STATS(vm);
+	vm->time += millitime(tstart, tend);
+
+	if (!gravity_vm_preempted(vm))
+			PRINT_STATS(vm);
```
InusualZ: ditto

sargun:
> The ability to be able to interrupt the VM (and step) will open us the possibility to attach a debugger to it.

The callback based method will allow for this as well, specifically, the runtime based callback.

sargun:
> If you don't have threads, in which to interrupt the VM, why not look into a yield mechanic, or something similar?
> Granted, that doesn't have the exact same functionality, unless you want to introduce a language level yield, but again, you would still differ in functionality.
> Why not a "debug/halt" opcode that halts the VM, also potentially reducing back pressure?

A couple things:
1. To me, this doesn't feel much better than coroutines. 
2. What would the C API look like? Would it unwind the stack? If so, we would still need the resume API.

sargun:
Is the fear of around the callback mechanism about performance?

sargun:
How about we allow people to define a macro instead of a full-on function. This way, if they want to use the flag, they can have a struct like:

```
struct wrapper_vm {
    bool preempted;
    gravity_vm vm;
}
```

And their macro can look like this:

```
static inline bool maybe_preempt(gravity_vm *vm) {
    struct wrapper_vm *wvm = container_of(vm, struct wrapper_vm, vm);
    return wvm->preempted;
}

#define MAYBE_PREEMPT maybe_preempt(vm)
// It an exercise left to the user to reset the state prior to calling resume.
```

marcobambini:
@sargun I'd like to have the ability to decide the pre-emption based on a runtime bool value instead of a macro that would require recompilation. I think that my latest proposal could be the best compromise between performance degradation (a bool comparison on each instruction) and the flexibility to interrupt the vm.

For debugging purpose some special bytecode instructions like BREAK and STEP would be the best option.

Please let me know what you think.

sargun:
What if we do something like have a config option for `PREEMPT_MODE`, and have a function `bool mark_preempt(gravity_vm *vm)` (where the bool returns true / false depending on if the VM was in a preemptible context) to tell a vm to yield if `#define PREEMPT_MODE internal` and alternatively, you can have `#define PREEMPT_MODE callback` and `#define PREEMPT_CB ...`?

marcobambini:
Hi @sargun I am going to re-open it because I'd like to use some of your ideas to implement a built-in debugger.

sargun:
Ah sounds good, I was just going through and cleaning out my github activity.

marcobambini:
Not planned

PR #97, opened 2017-03-18
String methods

hallzy:
Before you look at any of this, I just want to point out that the `test/string/storeat.gravity` fails, but I have checked it a few times and I don't see why it would fail. I checked the output of what I returned and it all looks good. So I am not sure if there is a bug in the unittest, or if I have missed something, but the result that I get from the test is what I expect

Since I don't have the time to go and finish up the date class right now, I thought I would contribute some string methods instead.

This will allow you to edit strings using the `[ ]`... ie, now `"Hello"[0]` will return a `H`... setting also works, so:

```swift
var s = "rain"
s[1] = "ipe" // now s is "ripe"
```
negative indices still work as well.

Also has an index() method that returns the index of the string that the matching search string is at:

```swift
"Hello World".index("Wo") // returns 6
```

count() method that counts the number of times that the search string occurs

string_repeat() which returns a new string that is a repeat of the original string... example:

```swift
"A".repeat_string(10) // returns "AAAAAAAAAA"
```

I initially wanted to use `repeat` but it was already being used by something in gravity... I believe the `repeat .. while` loop. I am open to better names, as I don't like the length of `repeat_string`

Then there is also `upper()` and `lower()` which just returns the string in uppercase or lowercase respectively.

There are a lot of other string methods that are prebuilt into other languages that would be worth exploring too that I have not done... like a split which would return a list where every element of the list is a part of the original string that is delimited by a specific string... basically the opposite of the join method gravity currently has.

### DIFF
--- a/src/runtime/gravity_core.c
+++ b/src/runtime/gravity_core.c
@@ -1450,6 +1450,171 @@ static bool operator_string_cmp (gravity_vm *vm, gravity_value_t *args, uint16_t
 	RETURN_VALUE(VALUE_FROM_INT(strcmp(s1->s, s2->s)), rindex);
 }
 
+static bool string_index (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+	#pragma unused(vm, nargs)
+	gravity_string_t *main_str     = VALUE_AS_STRING(GET_VALUE(0));
+
+	if (nargs != 2) {
+		RETURN_ERROR("String.index() expects 1 argument");
+	}
+	if (!VALUE_ISA_STRING(GET_VALUE(1))) {
+		RETURN_ERROR("String.index() expects a string as an argument");
+	}
+
+	gravity_string_t *str_to_index = VALUE_AS_STRING(GET_VALUE(1));
+
+	char *ptr;
+
+	// Search for the string
+	ptr = strstr(main_str->s, str_to_index->s);
+
+	// If it doesn't exist, return null
+	if (ptr == NULL) {
+		RETURN_VALUE(VALUE_FROM_NULL, rindex);
+	}
+	// Otherwise, return the difference, which is the index that the string starts at
+	else {
+		RETURN_VALUE(VALUE_FROM_INT(ptr-main_str->s), rindex);
+	}
+}
+
+static bool string_count (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+	#pragma unused(vm, nargs)
+	gravity_string_t *main_str     = VALUE_AS_STRING(GET_VALUE(0));
+
+	if (nargs != 2) {
+		RETURN_ERROR("String.count() expects 1 argument");
+	}
+	if (!VALUE_ISA_STRING(GET_VALUE(1))) {
+		RETURN_ERROR("String.count() expects a string as an argument");
+	}
+	gravity_string_t *str_to_count = VALUE_AS_STRING(GET_VALUE(1));
+
+	int j = 0;
+	int count = 0;
+
+	// Iterate through whole string
+	for (int i = 0; i < main_str->len; i++) {
+		if (main_str->s[i] == str_to_count->s[j]) {
+			// If the characters match and we are on the last character of the search
+			// string, then we have found a match
+			if (j == str_to_count->len - 1) {
+				count++;
+				j = 0;
+				continue;
+			}
+		}
+		// Reset if it isn't a match
+		else {
+			j = 0;
+			continue;
+		}
+		// Move forward in the search string if we found a match but we aren't
+		// finished checking all the characters of the search string yet
+		j++;
+	}
+
+	RETURN_VALUE(VALUE_FROM_INT(count), rindex);
+}
+
+
+static bool string_repeat (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+	#pragma unused(vm, nargs)
+	gravity_string_t *main_str = VALUE_AS_STRING(GET_VALUE(0));
+
+	if (nargs != 2) {
+		RETURN_ERROR("String.string_repeat() expects 1 argument");
+	}
+	if (!VALUE_ISA_INT(GET_VALUE(1))) {
+		RETURN_ERROR("String.string_repeat() expects an integer as an argument");
+	}
+	gravity_int_t times_to_repeat = VALUE_AS_INT(GET_VALUE(1));
+
+	if (times_to_repeat < 1) {
+		RETURN_ERROR("String.string_repeat() expects an integer >=1");
+	}
+
+	// Figure out the size of the array we need to make to hold the new string
+	int new_size = main_str->len*times_to_repeat;
+	char new_str[new_size];
+	strcpy(new_str, main_str->s);
+	for (int i = 0; i < times_to_repeat-1; i++) {
+		strcat(new_str, main_str->s);
+	}
+
+	RETURN_VALUE(VALUE_FROM_STRING(vm, new_str, strlen(new_str)), rindex);
+}
+
+static bool string_upper (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+	#pragma unused(vm, nargs)
+	gravity_string_t *main_str = VALUE_AS_STRING(GET_VALUE(0));
+
+	if (nargs != 1) {
+		RETURN_ERROR("String.upper() expects no argument");
+	}
+
+	for (int i = 0; i < main_str->len; i++) {
+		main_str->s[i] = toupper(main_str->s[i]);
+	}
+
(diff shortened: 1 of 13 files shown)


src/runtime/gravity_core.c
```diff
@@ -1450,6 +1450,171 @@ static bool operator_string_cmp (gravity_vm *vm, gravity_value_t *args, uint16_t
...
+	if (!VALUE_ISA_STRING(GET_VALUE(1))) {
+		RETURN_ERROR("String.index() expects a string as an argument");
+	}
+
+	gravity_string_t *str_to_index = VALUE_AS_STRING(GET_VALUE(1));
+
+	char *ptr;
+
+	// Search for the string
+	ptr = strstr(main_str->s, str_to_index->s);
```
InusualZ: Why you don't declare the variable here?

src/runtime/gravity_core.c
```diff
@@ -1450,6 +1450,171 @@ static bool operator_string_cmp (gravity_vm *vm, gravity_value_t *args, uint16_t
...
+
+	// If it doesn't exist, return null
+	if (ptr == NULL) {
+		RETURN_VALUE(VALUE_FROM_NULL, rindex);
+	}
+	// Otherwise, return the difference, which is the index that the string starts at
+	else {
+		RETURN_VALUE(VALUE_FROM_INT(ptr-main_str->s), rindex);
```
InusualZ: I don't know if this is part or the Gravity Formatting or what but, `ptr-main_str->s` In my opinion should be space between the minus operator like `ptr - main_str->s`. I had too look at the comment to be able to realize that he was doing subtraction.

src/runtime/gravity_core.c
```diff
@@ -1450,6 +1450,171 @@ static bool operator_string_cmp (gravity_vm *vm, gravity_value_t *args, uint16_t
 	RETURN_VALUE(VALUE_FROM_INT(strcmp(s1->s, s2->s)), rindex);
 }
 
+static bool string_index (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+	#pragma unused(vm, nargs)
+	gravity_string_t *main_str     = VALUE_AS_STRING(GET_VALUE(0));
```
InusualZ: space between the variable name and the equal op (`=`) should not be a thing

src/runtime/gravity_core.c
```diff
@@ -1450,6 +1450,171 @@ static bool operator_string_cmp (gravity_vm *vm, gravity_value_t *args, uint16_t
 	RETURN_VALUE(VALUE_FROM_INT(strcmp(s1->s, s2->s)), rindex);
 }
 
+static bool string_index (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+	#pragma unused(vm, nargs)
```
InusualZ: You should remove `nargs` from the `#pragma unused(vm, nargs)`. Same below

src/runtime/gravity_core.c
```diff
@@ -1450,6 +1450,171 @@ static bool operator_string_cmp (gravity_vm *vm, gravity_value_t *args, uint16_t
...
+	}
+	// Otherwise, return the difference, which is the index that the string starts at
+	else {
+		RETURN_VALUE(VALUE_FROM_INT(ptr-main_str->s), rindex);
+	}
+}
+
+static bool string_count (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+	#pragma unused(vm, nargs)
+	gravity_string_t *main_str     = VALUE_AS_STRING(GET_VALUE(0));
```
InusualZ: Again, should remove space between variable name and equal op

src/runtime/gravity_core.c
```diff
@@ -1450,6 +1450,171 @@ static bool operator_string_cmp (gravity_vm *vm, gravity_value_t *args, uint16_t
...
+		// finished checking all the characters of the search string yet
+		j++;
+	}
+
+	RETURN_VALUE(VALUE_FROM_INT(count), rindex);
+}
+
+
+static bool string_repeat (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+	#pragma unused(vm, nargs)
```
InusualZ: Should remove the `#pragma` since you are using those var. Same below

src/runtime/gravity_core.c
```diff
@@ -1450,6 +1450,171 @@ static bool operator_string_cmp (gravity_vm *vm, gravity_value_t *args, uint16_t
...
+
+	// Figure out the size of the array we need to make to hold the new string
+	int new_size = main_str->len*times_to_repeat;
+	char new_str[new_size];
+	strcpy(new_str, main_str->s);
+	for (int i = 0; i < times_to_repeat-1; i++) {
+		strcat(new_str, main_str->s);
+	}
+
+	RETURN_VALUE(VALUE_FROM_STRING(vm, new_str, strlen(new_str)), rindex);
```
InusualZ: you are calling `strlen`, but you already stored `new_str` length in `new_size` use it :)

src/runtime/gravity_core.c
```diff
@@ -1450,6 +1450,171 @@ static bool operator_string_cmp (gravity_vm *vm, gravity_value_t *args, uint16_t
...
+	gravity_string_t *string = VALUE_AS_STRING(GET_VALUE(0));
+	gravity_value_t value = GET_VALUE(1);
+	if (!VALUE_ISA_INT(value)) RETURN_ERROR("An integer index is required to access a string item.");
+	
+	register int32_t index = (int32_t)VALUE_AS_INT(value);
+	
+	if (index < 0) index = string->len + index;
+	if ((index < 0) || ((uint32_t)index >= string->len)) RETURN_ERROR("Out of bounds error: index %d beyond bounds 0...%d", index, string->len-1);
+
+	char c[2] = "\0";
```
InusualZ: Is this a leftover or something?

InusualZ:
I would be nice if in the function `String.upper/lower` you could specify an certain character that you want to be lower/upper. Another good function would be `explode(delimiter)` an example of what would do that function is:

```
"var e = Hello World".explode(' ')
System.print(e[0]); // Hello
System.print(e[1]); // World
```

marcobambini:
Thanks a lot @hallzy for your contribution.
I am going to merge your pull request now and then to take a look at why store at fails.
I'd like also to have repeat instead of string_repeat as method name.

marcobambini:
https://github.com/marcobambini/gravity/commit/112be515b5ef3b67011c7272e5a50ac3a1fcadc4 and https://github.com/marcobambini/gravity/commit/4795b2913b85d70dd506d743ef7cb254e875b7a7

hallzy:
@InusualZ Thanks for the suggestions. Looks like @marcobambini went and fixed them up as he went to change the `repeat_string` to work as `repeat`.

As for the explode, I agree. That would be nice to have. Either calling it explode or split.

The implode method you are talking about is already a feature of the List class, and is called join. I believe it is at the bottom of the docs for Lists.

I also like the idea of being able to specify the characters to make lower or uppercase. Maybe even do a bunch at once? ex `"hello world".upper(0, 6)` gives back `"Hello World"`.

I think I will quickly add that to the lower and upper actually.

src/runtime/gravity_core.c
hallzy: I am not sure specifically what you are referring to, but everything here looks like it should be there.

That index check is to allow negative indices to be used to start from the end of the string (it is code that is used elsewhere as well), and line 1589 is setting up a single character string... since replacing c[0] with a character (for example "x") makes `c == "x", due to the \0 null terminator.

src/runtime/gravity_core.c
hallzy: @marcobambini made this change in a commit referenced below

src/runtime/gravity_core.c
hallzy: @marcobambini made this change in a commit referenced below

src/runtime/gravity_core.c
hallzy: @marcobambini made this change in a commit referenced below

src/runtime/gravity_core.c
hallzy: @marcobambini made this change in a commit referenced below

src/runtime/gravity_core.c
hallzy: I changed this in an updated commit... thanks :)

src/runtime/gravity_core.c
hallzy: I changed this in an updated commit... thanks :)

Looking at it now I 100% agree, that was not a good decision on my part.

src/runtime/gravity_core.c
hallzy: I changed this in an updated commit... thanks :)

hallzy:
I have now actually implemented upper() and lower() to work as stated above... I just need to test it a bit more and try to make the code a bit nicer.

PR #98, opened 2017-03-18
Implemented Windows DLL build. Improved C++ include support.

brandon-ray:
On Windows ```make lib``` will now build ```gravity.dll``` for use in embedding Gravity in Windows applications.

Functions that are meant to be a part of the API need to be defined with ```GRAVITY_DLL``` so the compiler can export them properly.


marcobambini:
@brandon-ray thanks a lot for your changes. May I ask you to rename the GRAVITY_DLL macro to GRAVITY_API? The reason is that _DLL seems a Windows only feature while a more generic _API suffix will give us the ability to override linkage features on other platforms too.

On the Makefile LIBTARGET for Linux should be libgravity.so and for MacOS should be libgravity.dylib
Thanks!

brandon-ray:
Ah good idea, I will do so.

marcobambini:
Thanks a lot!

wqweto:
The way `GRAVITY_API` is defined it is not possible to build a DLL using stdcall calling convention. Most windows languages can use win32 API (which is stdcall) but some have difficulties with cdecl per se.

The API macro has to be something like `GRAVITY_API(type)` with the type being a param, to be able to plug an optional `__stdcall` modifier after `type` in function declaration.

marcobambini:
@wqweto can you please give me a concrete example?

PR #100, opened 2017-03-18
String.upper() and .lower() shouldn't modify the string directly

hallzy:
I discovered while looking over my code that the upper() and lower() methods themselves actually modify the base string. For example:

```swift
var str = "HeLLo"
str.Upper()
System.print(str) // prints out "HELLO"
```

This was not my intention when I made these methods. My intention was:

```swift
var str = "HeLLo"
var ret = str.Upper()
System.print(str) // prints out "HeLLo"
System.print(ret) // prints out "HELLO"
```

so I have added to the testcases, and modified the code to resolve that issue. I apologize for this error.


marcobambini:
Thanks @hallzy!

PR #101, opened 2017-03-19
Add args to upper lower

hallzy:
Just want to say not to merge this yet. I just pushed so we can have a look at it. Before a merge happens I would prefer to squash the commits into one, but until I know this is good, I would rather leave it like this so I can go back if needed.

Basically, now the upper() and lower() methods take arguments that are either strings or integers. And it can take multiple arguments

```swift
str.upper(0, "h", "What", -2)
```

The above example will uppercase the characters of the string marked by index 0, and -2... and it also uppercases the parts of the string that match "h" and "What".

The functions got really quite long, so I created a new separate function that gets executed from each of the upper() and lower() methods which does most of the work. That function takes two function pointers, to allow me to use the same function for both lower() and upper() (one function pointer is either `isupper` or `islower`, and the other is `toupper` or `tolower`.

I added a lot of test cases as well just to try and make sure that I cover a lot of different scenarios.


marcobambini:
Hi @hallzy I can understand the value of the integer argument but the string type argument is really useful?
Seems like an huge amount of code for a feature probably rarely used.

hallzy:
Perhaps. Not having to worry about strings certainly would simplify it a lot. It was kind of fun to do though :)

If you think it would be best to omit it, I can remove it and change the documentation and unittests accordingly, and I will save the code for supporting strings locally just in case it becomes something to revisit.

marcobambini:
Well done @hallzy, thanks a lot.

PR #106, opened 2017-03-22
Added a random() method for System

hallzy:
```swift
System.random(-5, 10) // returns a random integer between -5 and 10 inclusively
```

### DIFF
--- a/docs/system.html
+++ b/docs/system.html
@@ -139,6 +139,14 @@
 	$ echo $?
 				</code></pre>
 				
+				<h4 class="section-h4">Random Number</h4>
+				<p>System has a random number generator:</p>
+				<pre><code class="swift">
+	func main() {
+		System.random(1, 10) // Returns an integer between 1 and 10 inclusively
+	}
+				</code></pre>
+
          	</div>
          	<!-- END CONTENT -->
          	
--- a/src/runtime/gravity_core.c
+++ b/src/runtime/gravity_core.c
@@ -8,6 +8,8 @@
 
 #include <inttypes.h>
 #include <math.h>
+#include <time.h>
+#include <stdlib.h>
 #include "gravity_core.h"
 #include "gravity_hash.h"
 #include "gravity_value.h"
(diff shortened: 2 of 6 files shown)


marcobambini:
Thanks @hallzy. As soon as  I'll finalize the opt subsystem we'll move the rand method to the more appropriate Math class.

hallzy:
Okay sounds good. I wasn't sure if it would be better suited in a math class or a system. Javascript seems to have it as part of their math class, but C seems to just include it in stdlib rather than their math header for instance.

Math seems like a descent place to put it though

hallzy:
I was not aware of this. I have always done it that way. How would you generate a random number in C?

kazzkiq:
Cool method, but wouldn't this make more sense inside a math-related class?

Example:

```js
// All mathematical methods goes inside this "Math" class
Math.random(-5, 10);
```

hallzy:
@marcobambini mentioned this earlier in this PR. I think it could probably work in either, but Math certainly makes sense, and I would not object to this change.

I believe he is going to merge it into a Math class once he finalizes the opt subsystem (basically, he is adding a way for us to more easily add new classes in a file separate from the gravity_core... so once he has done that we will be moving it)

kazzkiq:
@hallzy Oh, sorry, completely missed that comment. 😄

hallzy:
np :)

marcobambini:
@hallzy instead of the old and insure stand/rand functions we should use arc4random (available in stdlib.h). This can helps http://stackoverflow.com/questions/3420581/how-to-select-range-of-values-when-using-arc4random

hallzy:
I am unable to get it to work. It doesn't seem to be in stdlib. I get a compilation warning for an implicit declaration, and a compilation error for the unittest target.

When I look up anything about it I get more stuff related to Swift and Obj-C than actual C.

brandon-ray:
Should this function be named differently to show it only handles integers? Something like ```randomInt``` or similar. Then we can have other random value functions like ```randomFloat`` and then the "regular" random function ```random``` could return a value between 0 and 1 like other languages.

It would seem odd to me to only support random integers.

marcobambini:
Good point @brandon-ray and based on your observation I'd suggest to move the random method from System meta to Int meta class so user will be able to write:
Int.random(...)

brandon-ray:
Ah that is an even better idea, then we can have ```Float.random(...)```, ```Bool.random()```, etc.

hallzy:
I like that idea better than just a general system or math method.

If someone else wants to make the change, go ahead... I don't think I am going to have much free time for at least the next few days

PR #107, opened 2017-03-22
Apply method does not fill called function _args variable

parro-it:
While changing [gravity-func](https://github.com/parro-it/gravity-func) to introduce the new apply method, I found what I think could be a bug in `apply` method.

`apply` correctly pass arguments to the called functions, but the `_args` dynamic argument only contains the first argument.

In this PR I add a failing test that check _args variable, and some log. 

The PR is not meant to be merged...


src/runtime/gravity_core.c
```diff
@@ -1038,7 +1038,11 @@ static bool closure_apply (gravity_vm *vm, gravity_value_t *args, uint16_t nargs
 	gravity_value_t self_value = GET_VALUE(1);
 	gravity_list_t *list = VALUE_AS_LIST(GET_VALUE(2));
 	
-	gravity_vm_runclosure(vm, closure, self_value, list->array.p, (uint16_t)marray_size(list->array));
+	uint16_t argsCount = (uint16_t)marray_size(list->array);
+
+	printf("apply args size: %d\n", argsCount);
+	printf("args: %ld %ld\n", list->array.p[0].n, list->array.p[1].n);
+	gravity_vm_runclosure(vm, closure, self_value, list->array.p, argsCount + 1);
```
parro-it: Running the new test, these lines output

```
apply args size: 2
args: 1 2
```
That seems correct...

parro-it: Here, by adding +1 to argsCount, the test pass, all arguments are copied in _args variable.
But seems incorrect and fragile to add 1...

parro-it: If I don't add it, `_args[0] == 1` and `_args.count == 1`

src/runtime/gravity_core.c
```diff
@@ -1038,7 +1038,11 @@ static bool closure_apply (gravity_vm *vm, gravity_value_t *args, uint16_t nargs
 	gravity_value_t self_value = GET_VALUE(1);
 	gravity_list_t *list = VALUE_AS_LIST(GET_VALUE(2));
 	
-	gravity_vm_runclosure(vm, closure, self_value, list->array.p, (uint16_t)marray_size(list->array));
+	uint16_t argsCount = (uint16_t)marray_size(list->array);
+
+	printf("apply args size: %d\n", argsCount);
+	printf("args: %ld %ld\n", list->array.p[0].n, list->array.p[1].n);
```
InusualZ: I don't think these printf should end up in the final product

parro-it: This is not meant to be merged but to showcase a bug... 😀

marcobambini:
Thanks a lot @parro-it, fixed by https://github.com/marcobambini/gravity/commit/4149aeece34638c7030073ebe684cecd96bf04c5 and unit test added.

parro-it:
👏👏

PR #108, opened 2017-03-23
Fix JSON pretty printing when not inside an object and a key is given.

boardwalk:
I was looking at the compiler's output and I saw something like this:

```
        "nlocal" : 0,
        "ntemp" : 1,
        "nup" : 0,
        "purity" : 0.500000,
        "bytecode" : "200400003C040001C40400023C040003",
        "pool" : [
{
                "type" : "class",
                "identifier" : "Vector",
                "nivar" : 3,
                "init" : {
                    "type" : "function",
                    "identifier" : "init",
```

This fixes the indentation of that curly brace and makes `json_begin_array` robust against passing a key when not in an object like `json_begin_object`.

### DIFF
--- a/src/utils/gravity_json.c
+++ b/src/utils/gravity_json.c
@@ -155,16 +155,18 @@ static void json_write_escaped (json_t *json, const char *buffer, size_t len, bo
 
 
 void json_begin_object (json_t *json, const char *key) {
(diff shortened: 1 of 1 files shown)


marcobambini:
Thanks a lot @boardwalk!

PR #110, opened 2017-03-27
Changed System.random() to Int.random()

hallzy:
I moved the System.random() method to Int.random() as discussed in #106.

Have not done any floats or bools etc. Also, the rand() function is still being used as I was still not able to get that arc4random function working.


marcobambini:
Thanks a lot @hallzy !

PR #112, opened 2017-03-30
Added a String.split() method.

hallzy:
Acts as a reverse to the List.join()

Only takes one (string) argument.

### DIFF
--- a/docs/types.html
+++ b/docs/types.html
@@ -124,6 +124,12 @@
 	// upper() and lower() can both take multiple integer arguments
 	n = b.upper(1, -1)     // n is now "HEllo WorlD"
 
+	// split() is the opposite of a list join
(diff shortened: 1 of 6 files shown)


marcobambini:
Thanks @hallzy but there are some memory leaks in your code.
You use strdup but you never free memory.

You can avoid the **strdup** calls just using **strstr** over the original string and then use (**p-string->s**) to obtain the length on the substring (without the need to zero terminate it). Then in order to create a **gravity_string_t** you can use **VALUE_FROM_STRING** passing **p** as pointer and (**p-string->s**) as length.

marcobambini:
I am going to merge your pull request and then improve the string_split.

hallzy:
Yup it would appear that your solution was much better. Thnx for fixing it up.

PR #116, opened 2017-04-02
Added a String.substr() method

hallzy:
added a substring method that takes either 1 or 2 arguments.

if one argument is provided, it takes the substring defined by that index of the string, to the end of the string, otherwise it is the string defined by two given indices.

For example:

```swift
var str = "Hello, World!"
return str.substr(7, 11) // "World"... or str.substr(7, -2)
```

Not sure if someone else has a better name for it, but I am open to suggestions if substr() is not appealing.

It also seems that some languages specify the ending index of the character after the last character you want, while this stops the substring at the character you specify...

For instance, in JS that substring command I gave above would return "Worl" instead of "World".

I have also seen some languages take the substring length as the second argument and that is a possibility as well, instead of what this implementation gives.

Perhaps this needs further discussion on if the name needs to be changed, or we want to change the meaning of the 2nd index

### DIFF
--- a/docs/types.html
+++ b/docs/types.html
@@ -138,6 +138,10 @@
 	// and retrieve those characters
 	n = a[6]                // n is now "W"
 	n = a[-7]               // n is now "o"
+	n = a[0...4]            // n is now "Hello"
+	n = a[-5...-1]          // n is now "World"
+	n = a[-5...10]          // n is now "World"
+	n = a[-1...-5]          // n is now "dlroW"
 			</code></pre>
 			
 			<h4 class="section-h4">Bool</h4>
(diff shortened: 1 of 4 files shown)


marcobambini:
I'd suggest to just extends the string_loadat implementation (it now accepts an integer index only) with range support. In this way you can use a syntax like:
`string[1...3]`
to get a substring based on a range index.

Take a look at https://github.com/munificent/wren/blob/master/test/core/string/subscript_range.wren for some examples.

hallzy:
That's a good idea... I will have a look later on

hallzy:
Okay, I have modified the string loadat to handle ranges...

I also did what wren did, and if the range is in reverse order, it returns a reversed string as well. And negative indices still work.

marcobambini:
Very well done @hallzy I really appreciate your help!

marcobambini:
I wondering if the three costly **strlen(original)** can be replaced by **string->len**.

hallzy:
It could be replaced with `string->len` for sure. Or `uint32_t original_len = string->len` and use `original_len` instead.

PR #122, opened 2017-04-06
Fixed a compilation warning: %ld vs %lld

hallzy:


marcobambini:
@hallzy inst->n is an int64_t number so the right way is to use PRId64.
I'll commit the fix in few hours.

PR #162, opened 2017-05-25
Build on OpenBSD.

ibara:
As it says, in preparation for adding gravity to the official OpenBSD packages repository.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,10 @@ else
 		# MacOS
 		LIBTARGET = libgravity.dylib
 		LDFLAGS = -lm 
+	else ifeq ($(UNAME_S),OpenBSD)
+		# OpenBSD
+		# LIBTARGET = libgravity.so (not used)
+		LDFLAGS = -lm
 	else
 		# Linux
 		LIBTARGET = libgravity.so
--- a/api/Makefile
+++ b/api/Makefile
@@ -22,6 +22,9 @@ else
 	ifeq ($(UNAME_S),Darwin)
 		# MacOS
 		LDFLAGS = -lm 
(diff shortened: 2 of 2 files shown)


ibara:
Also, here is a link to the test log for OpenBSD. There appears to be 8 test failures.
https://devio.us/~bcallah/gravity-tests-openbsd.log

marcobambini:
Hi @ibara there are 8 test failures because of a recent changes to the unused module code.
You can safely ignore all of them. I'll soon update the unit test.

marcobambini:
@ibara, modules related unittest are now disabled (https://github.com/marcobambini/gravity/commit/63a726b5a88e6c06f55ce6b93eb207daef4724f0).
Do you still have one failure?
If yes, which one?

ibara:
I have one failure now.
```
865     Test file: include_test.gravity
        Test path: test/unittest/include_test.gravity
        Test #include macro.
        SYNTAX ERROR on 0 (7,11): Unable to load file test.inc.
        FAILURE
```

marcobambini:
It calls callback_read in unittest.c and the failure should be in file_buildpath (in gravity_utils.c).
The issue should be a wrong file path computed from the current test folder.

Can you please just execute that unittest and then add a printf to print full_path in file_buildpath?

ibara:
Hi. I did the above and this is what I get:
```
867     Test file: include_test.gravity
        Test path: test/unittest/include_test.gravity
        Test #include macro.
test/test.inc
        SYNTAX ERROR on 0 (7,11): Unable to load file test.inc.
        FAILURE
```
If I change the include line in include_test.gravity to:
`#include "unittest/test.inc"`
then the test passes.

PR #174, opened 2017-07-22
Added -q option to cli to be quiet result and execution time output

MakeNowJust:
It is useful to pass gravity's output to other command by pipe like `gravity -q script.gravity | wc -l`.


marcobambini:
Thank you so much for the feature.

PR #178, opened 2017-08-19
Added some functionality to Math class

hallzy:
I have added a cube root function (cbrt) along with a gcf and lcm, and in addition to the log function that already existed, I added a log10 for base 10 logs, and a logx for variable based logs (ie, "Math.logx(2,4)" is a base 2 log of 4 and equates to 2).

I also fixed the Math.random() function as it always resulted in a 0 (the random value just needed to be casted to a float).

I also added a degrees and radians property to the Int and Float types so that you can say "30.radians" to convert 30 degrees into radians for example.

And finally, the current gitignore was ignoring the "test/unittest" folder because of the line "unittest" that was meant to block the unittest executable. So I added an exception to the gitignore for that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -283,6 +283,7 @@ paket-files/
 *.hex
 gravity
 unittest
+!test/unittest
 
 # Debug files
 *.dSYM/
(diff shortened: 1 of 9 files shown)


marcobambini:
Very well done, thanks!

PR #180, opened 2017-08-20
Added a Math class docs file

hallzy:
Noticed that there wasn't a documentation file for the new Math class, so I went ahead and added one quickly.


marcobambini:
Very helpful, thanks a lot!

PR #181, opened 2017-08-20
Updated math doc after the new math functions were merged

hallzy:
This is the updated math docs after my additions to the math class were just merged.

### DIFF
--- a/docs/math.html
+++ b/docs/math.html
@@ -100,16 +100,18 @@
 							<h4 class="section-h4">Trig Functions</h4>
 							<p>The Math class also contains several Trigonometric Functions. All
 									values that represent angles are in radians for these methods, and they
-									all expect radians for inputs.</p>
+									all expect radians for inputs. You can use the ".radians" and
+							".degrees" properties of the Int and Float class to do conversions.</p>
 							<pre><code class="swift">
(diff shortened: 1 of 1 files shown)


marcobambini:
Not sure about the conflict to be resolved.
Any hint?

hallzy:
I'll have a look tonight when I have a computer I can sit at.

hallzy:
actually just looked at it but I can't really fix it here. Keep both of the lines if you want to do it, otherwise I will do it tonight.

hallzy:
Not sure what happened but it looks like some stuff wasn't merged, such as the cbrt, log10 and logx stuff. This page says that this has been closed with unmerged commits and based on the states of the files it looks like it the changes that are missing should now just go in.

marcobambini:
@hallzy not sure what happened.
I just kept both the the lines as you suggested.
Can you please take a look at this issue?

hallzy:
Sure not a problem. I will do it in the morning or afternoon though as it's 2:30am here.

hallzy:
Actually it looks like it can be merged the way it is now.

marcobambini:
Thanks a lot, I really appreciate.

PR #183, opened 2017-08-26
Added a string iterator and loop function.

hallzy:
This allows a user to loop through a string character by character using
the for-in loop, or the loop() method.

### DIFF
--- a/src/runtime/gravity_core.c
+++ b/src/runtime/gravity_core.c
@@ -2058,6 +2058,62 @@ static bool string_split (gravity_vm *vm, gravity_value_t *args, uint16_t nargs,
 	RETURN_VALUE(VALUE_FROM_OBJECT(list), rindex);
 }
 
+static bool string_loop (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+	if (nargs < 2) RETURN_ERROR("Incorrect number of arguments.");
+	if (!VALUE_ISA_CLOSURE(GET_VALUE(1))) RETURN_ERROR("Argument must be a Closure.");
+
+	gravity_closure_t *closure = VALUE_AS_CLOSURE(GET_VALUE(1));	// closure to execute
+	gravity_value_t value = GET_VALUE(0);							// self parameter
+	gravity_string_t *string = VALUE_AS_STRING(value);
+	char *str = string->s;
+	register gravity_int_t n = string->len;  // Times to execute the loop
+	register gravity_int_t i = 0;
+
+	nanotime_t t1 = nanotime();
+	while (i < n) {
+		gravity_value_t v_str = VALUE_FROM_STRING(vm, str + i, 1);
+		if (!gravity_vm_runclosure(vm, closure, value, &v_str, 1)) return false;
+		++i;
+	}
+	nanotime_t t2 = nanotime();
+	RETURN_VALUE(VALUE_FROM_INT(t2-t1), rindex);
+}
+
+static bool string_iterator (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+	#pragma unused(vm, nargs)
+	gravity_string_t *string = VALUE_AS_STRING(GET_VALUE(0));
+
+	// check for empty string first
+	if (string->len == 0) RETURN_VALUE(VALUE_FROM_FALSE, rindex);
+
+	// check for start of iteration
+	if (VALUE_ISA_NULL(GET_VALUE(1))) RETURN_VALUE(VALUE_FROM_INT(0), rindex);
+
+	// extract value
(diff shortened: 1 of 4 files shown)


marcobambini:
Very well done thanks.
Not sure to understand why in the string_loop function you first create a temporary gravity_list_t *list, seems like a waste of resource... am I missing something?

src/runtime/gravity_core.c
```diff
@@ -2058,6 +2058,70 @@ static bool string_split (gravity_vm *vm, gravity_value_t *args, uint16_t nargs,
...
+	char *str = string->s;
+	register gravity_int_t n = string->len;  // Times to execute the loop
+	register gravity_int_t i = 0;
+
+	gravity_list_t *list = gravity_list_new(vm, 0);
+	for (int i = 0; i <= n; ++i) {
+		// A litte trick to make a 1 character string so that VALUE_FROM_STRING will
+		// accept it.
+		char c[1] = {str[i]};
+		marray_push(gravity_value_t, list->array, VALUE_FROM_STRING(vm, c, 1));
```
InusualZ: Instead of creating an array of 1 you could just create one char and get the pointer.
```
char c = str[i];
marray_push(gravity_value_t, list->array, VALUE_FROM_STRING(vm, &c, 1));
```

Or even better with an little bit of pointer arithmetic you could accomplish the same thing without creating an temporary variable. Although I'm not very familiarized with C so I don't know if it's valid or a good thing to do.

```
marray_push(gravity_value_t, list->array, VALUE_FROM_STRING(vm, str + i, 1));
```

src/runtime/gravity_core.c
```diff
@@ -2058,6 +2058,70 @@ static bool string_split (gravity_vm *vm, gravity_value_t *args, uint16_t nargs,
...
+static bool string_loop (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+	if (nargs < 2) RETURN_ERROR("Incorrect number of arguments.");
+	if (!VALUE_ISA_CLOSURE(GET_VALUE(1))) RETURN_ERROR("Argument must be a Closure.");
+
+	gravity_closure_t *closure = VALUE_AS_CLOSURE(GET_VALUE(1));	// closure to execute
+	gravity_value_t value = GET_VALUE(0);							// self parameter
+	gravity_string_t *string = VALUE_AS_STRING(value);
+	char *str = string->s;
+	register gravity_int_t n = string->len;  // Times to execute the loop
+	register gravity_int_t i = 0;
```
InusualZ: I'm a little bit confused about this. You have this variable named `i` and in the for loop you create an scoped variable name `i`. Did that even compile? Consider renaming one of them, for code readability

src/runtime/gravity_core.c
hallzy: Nice, didn't think of that one. Your second one does work.

src/runtime/gravity_core.c
hallzy: Wow, you are absolutely right. It did actually compile.

Since this can be a learning moment for all since this seems like something that shouldn't compile, I looked at it and apparently (at least in some cases) it is valid, but you will only have access to the `i` variable that was declared in the for loop's scope.

Consider this c example:

```c
#include <stdio.h>

int main() {
  int i = 5;
  for (int i = 0; i < 10; i++) {
    printf("%d", i);
  }
  printf("\n%d\n", i);
  return 0;
}
```

The output is:

```
0123456789
5
```

As the loop was only using the `i` that was in the for loop declaration, but when I exited the for loop that for loop `i` variable was no longer in scope, so it used the other `i`.

Very strange. But while it does work I agree with you it is confusing, so I will change it.

hallzy:
@InusualZ I have made both of you recommendations and the PR has been updated.

@marcobambini I am using the list so that I can add each character to the list one at a time. I could not seem to get it to work otherwise.

If I used:

```swift
if (!gravity_vm_runclosure(vm, closure, value, &VALUE_FROM_STRING(vm, str + i, 1), 1)) return false;
```

instead of:

```swift
if (!gravity_vm_runclosure(vm, closure, value, &marray_get(list->array, i), 1)) return false;
```

It produces an error `lvalue required as unary '&' operand`

It has occurred to me now while writing this that doing:

```swift
gravity_value_t v = VALUE_FROM_STRING(vm, str + i, 1);
if (!gravity_vm_runclosure(vm, closure, value, &v, 1)) return false;
```

Does actually work though, and is probably better than defining a new list, but at the time of writing it nothing I tried worked and I didn't think of that.

I have updated the code with this new change, but if you have a better way of doing it I am all ears.

marcobambini:
Thanks @hallzy I have now merged your pull request.
Your code will probably not work with UTF-8 strings, I should probably first add a new utf-8 character extraction function in utils.c

PR #184, opened 2017-09-02
Remove trailing whitespace in many files.

neysofu:
As common knowledge has it, trailing whitespace is bad. It creates commit noise, it is inconsistent, and makes editing files harder with editors that rely on hotkeys and keyboard commands.

I have deleted trailing whitespace in many files. I made sure to do so only when it made sense to me and when I was sure it wouldn't break anything.


marcobambini:
@neysofu can you please update your PR to have no conflicts with the latest version?
Thanks.

neysofu:
Working on it right now. It will be ready in a few hours.

neysofu:
Ready to merge.

marcobambini:
Thanks @neysofu!

Issue #104, opened 2017-03-22
Built in Sort


hallzy:
It might be useful to have some kind of built in sorting algorithm for things like lists (and maybe a reverse):

```swift
var list = [2, 4, 1, 5, 3]
list.sort() // returns [1, 2, 3, 4, 5]
list.sort().reverse() // returns [5, 4, 3, 2, 1]
```

if this is something that may make it's way into gravity it is probably worth thinking about what would happen in the event that a list contains numbers and strings, or if we want other things to be sorted as well, such as maybe a list of objects sorted by a specific field.

Also what sorting algorithm would be used. I have heard that for example, python uses Timsort.

marcobambini:
Yes I agree that it would be really useful.

Things are easy when the list contains elements of the same type but became a little obscure when different types are contained. A list can contain numbers, strings, functions, classes and any kind of objects.

A simple algorithm would be to manage the number/string case in the traditional way and sort any other object by its hash.

kngu9:
The best general sorting algorithm for arbitrary data would be quicksort, which has O(n^2) but is theta(nlogn). 

@marcobambini The best way to do this is to make a callback function, so the developer can decide how to convert it.

`
func sort_func(a, b) {
    return a > b
}

var list = [1, 2, 3, 'a', 'b', 'c']
list.sort(sort_func)
`

neysofu:
See [this](https://stackoverflow.com/questions/7770230/comparison-between-timsort-and-quicksort) discussion for a comparison of Timsort and Quicksort.

I also think that allowing implicit ordering between types is a bad idea and it would be better to pass an explicit ordering function.

marcobambini:
Anyone is interested in working on this feature?

MatanSilver:
I've got a working reverse function and I can take a crack at a sort too if you like. I'm a bit new to writing for compilers but I can give it a shot...

marcobambini:
@MatanSilver your contribution would be more than welcomed :-)

MatanSilver:
Presumably we want sort and reverse to return new list objects and not modify the ones they are called on?

marcobambini:
That's a good question. Swift uses sort() to modify the original array while sorted() returns a new array:  [https://developer.apple.com/documentation/swift/array/1688499-sort](https://developer.apple.com/documentation/swift/array/1688499-sort) I think we can follow the same behaviour.

MatanSilver:
[...]

PR #187, opened 2017-10-15
Built in sort and reverse

MatanSilver:
This idea was originally floated in issue #104. Four list class methods have been added here--reverse(), reversed(), sort(), and sorted(). Sort() and sorted() both take a "predicate" argument, which is a function (closure) that takes two arguments and returns a "truthy" value. The sort functions use an implementation of quicksort, though Timsort was also suggested as a possibility.

Unittests have been added to make sure that reverse() and sort() modify the original list, while reversed() and sorted() return a new list and leave the original unmodified.

I'd appreciate any feedback!

### DIFF
--- a/src/runtime/gravity_core.c
+++ b/src/runtime/gravity_core.c
@@ -957,6 +957,73 @@ static bool list_reversed (gravity_vm *vm, gravity_value_t *args, uint16_t nargs
   RETURN_VALUE(VALUE_FROM_OBJECT(newlist), rindex);
 }
 
+static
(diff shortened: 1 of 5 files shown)


marcobambini:
Very well done @MatanSilver, I really like the implementation.
Please remember to add yourself to the CONTRIBUTORS file too.

marcobambini:
@MatanSilver seems like there are some issue that should be addressed https://github.com/marcobambini/gravity/issues/188

MatanSilver:
Thanks, I've fixed it as far as I can tell but I'll write a bunch more test cases to stress the implementation a bit more before submitting another pull request

Issue #194, opened 2017-10-19
Native map, reduce, and filter


MatanSilver:
Modeling off of https://useyourloaf.com/blog/swift-guide-to-map-filter-reduce/

MatanSilver:
Swift uses a named argument  "combine" in reduce, but has unnamed arguments for map and filter. Incidentally I noticed Swift uses a named argument "by" for the sort function. Should we just settle on unnamed arguments for these functions (It seems gravity only has unnamed arguments at the moment)?

marcobambini:
In Gravity unnamed arguments can be processed using the special __args[] array (I can introduce a syntactic sugar for the $n notation). The combine argument in reduce is just a closure that can be used as is. Anyway, I would just settle on unnamed arguments for these functions.

MatanSilver:
Also it seems like Swift doesn't have the same "reverse" vs "reversed" or "sort" vs "sorted" semantics for persistence on its map and filter (though I may have missed it in my searches). I think map should probably just return a new array since it could change the type in the list container which could make the type system confusing when doing that operation. Filter though is less obvious, and I was wondering if you had thoughts on what filter and friends should do in this regard (i.e maybe make both "filter" and "filtered")? Thanks

marcobambini:
Even if Swift does not have "reverse" vs "reversed" or "sort" vs "sorted" semantics I like the idea to be more expressive in Gravity. I agree with you that map should always return a new array. filter should always return a new array.

YurySolovyov:
not sure how relevant it is, but I had some stuff implemented here: https://github.com/YurySolovyov/lo-gravity

marcobambini:
Thanks @YurySolovyov for the very useful functions.
Matan added some c native functions so I am going to close this issue and I'll add a link to your repo into the main Wiki section: [https://github.com/marcobambini/gravity/wiki](https://github.com/marcobambini/gravity/wiki)

YurySolovyov:
I wish there was a way to implement them self-hosted inside native gravity package, so they are available for everyone right away.

marcobambini:
@YurySolovyov I'd like to have a Gravity package manager too.
Let me think a bit more about the best way to implement it.

YurySolovyov:
By "native package" I meant a way to implement them as part of gravity runtime, but without writing C code. Sorry for confusion.
This actually was my initial intent, but I realized there was no way to implement these without writing C, so I created `lo-gravity` instead. 

[...]

PR #196, opened 2017-10-22
Map reduce filter

MatanSilver:
This is my implementation of map, filter, and reduce, as originally suggested by @hallzy and outlined in #194 . I've added unit tests that I believe show that the methods work as expected. There are some potential optimizations to be made. For example, right now filter is optimized for speed, not memory. If memory is an issue, we can try to make sure that we only allocate the right size for the filtered array (right now it will almost always be oversized).


marcobambini:
To be more tested but thanks a lot @MatanSilver for your code.

Issue #198, opened 2017-12-02
Suggestion: For windows mingw-make, add CC variable in Makefile


shawnye:
Without `CC=gcc` in Makefile, error occurs when do `mingw-make`:

<pre>
process_begin: CreateProcess(NULL, cc...)failed. 
make (e=2):
</pre>

marcobambini:
Thanks @shawnye.
Fixed by https://github.com/marcobambini/gravity/commit/b9f36f6a8a88e8cff383f45e36c31779271cb58b

PR #203, opened 2018-02-13
Reduce CC = to CC ?= in Makefile

ibara:
Some packaging tools (e.g. OpenBSD ports) predefine CC, since our default compiler is clang, not gcc. This allows systems that don't have CC defined to use gcc, and those that do use their preselected compiler. As-is, your invocation of CC=gcc clobbers any predefined CC, and this breaks the OpenBSD ports build without additional patching on our end.


marcobambini:
Thanks @ibara I added that line because of https://github.com/marcobambini/gravity/issues/198

ibara:
No worries--this should make everyone happy now.

PR #208, opened 2018-03-26
Added unittest and !test/unittest to .gitignore

hallzy:
Without those lines, the unittest executable shows up in a git status as
untracked as opposed to be ignored.

These lines used to be in the gitignore but were removed in d7b73ef19dc351f06b3dbd5960766728e38a1fd3#diff-a084b794bc0759e7a6b77810e01874f2 though I am not sure why. If there was a reason for it can someone let me know.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -282,6 +282,8 @@ paket-files/
 *.x86_64
 *.hex
 gravity
+unittest
+!test/unittest
 
 # Debug files
 *.dSYM/


marcobambini:
@hallzy why unittest should be ignored?

hallzy:
it is just that if you don't ignore it it comes up in the git status:

```
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	unittest

nothing added to commit but untracked files present (use "git add" to track)
```

Also, the `gravity` executable is still in the gitignore so for consistency it makes some sense...

Mostly it is just a pet peeve of mine... I like seeing clean statuses :) Obviously it doesn't add to the code so if you deem it unnecessary that is fine. Just thought that I would point out that it used to be in the gitgnore.

marcobambini:
```
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.

nothing to commit, working tree clean
```

I wonder why I don't see any untracked file.
Please note that unittest should be tracked like any other file and not ignored or untracked.

hallzy:
well that's interesting. Just to be clear, I am talking about the `unittest` executable that is built with `make` so if you haven't built it yet or you just cleaned it that makes sense but otherwise I am kind of confused.

Also, the `gravity` executable that is built with `make` is ignored in the gitignore as well and that `gravity` executable is also not tracked (it is not in the source files on github for example).

marcobambini:
Ah you mean the executable!
Oh yes sure it should be added to gitignore, thanks!

PR #213, opened 2018-04-05
Added documentation for the map, filter, and reduce methods for lists

hallzy:
I noticed that while there was a map, filter, and reduce method for the lists class, there was no documentation for it in the docs, so I thought I would add it.


marcobambini:
Thanks @hallzy I really appreciate your help!

PR #214, opened 2018-04-05
split('') will now split a string at every character.

hallzy:
Previously, this was an undefined case that would result in an infinite
loop. Now it acts the same as in JavaScript, and it returns a list
where every element is 1 character of the string.

This is useful for applying list methods to strings such as map, reduce,
filter, contains etc


marcobambini:
Thanks @hallzy that's very useful!

PR #243, opened 2018-08-03
Windows support

stevefan1999:
This PR allows Windows to build Gravity under Visual Studio using MSVC >= 2015, it is separated into 3 parts, (lib)gravity for static library to rely on, gravity.exe for CLI and gravity.dll for shared library.


marcobambini:
Hi @stevefan1999 and thanks for the contribution, I really like it, butI have few questions/requests:

1. #pragma unused is not available on Window (and more generally on non clang compilers), please leave it as is as I plan to introduce a more general macro #define UNUSED(x) (void)(x)

2. in gravity_vm.c what is MSVC bug: designated initializer was prematurely evaluated ?

3. in gravity_value.c why have you renamed src to srcInstance? (also please note that I do not use camel case in Gravity)

Everything else looks pretty good to me.
Thanks again.

stevefan1999:
@marcobambini 
1. OK🤔
2. The macro expansion in `VALUE_FROM_OBJECT` contains a designed initializer and it accesses the `isa` field of the `obj` macro argument. I'm also confused at this that why there is an access violation to a null pointer while there is already a guard before that...It is a bug on MSVC that it always initialized the struct fields even if the branch is not yet determined...
3. The name is inconsistent with the declaration, I see that `instance` is already named so I change it to `srcInstance` instead.

marcobambini:
1. It turns out that I already added all the necessary macros (please take a look at gravity_macros.h:16)
2. Very good catch thanks. Do you know if the bug affects all MSVC version or just a particular version?
3. I fixed the internal names with my latest commit

marcobambini:
@stevefan1999 I'd like to include your changes, can you please re-sync your pull with the latest master version? Thanks!

marcobambini:
Thanks!

stevefan1999:
@marcobambini Regarding to a previous comment you made, I think the bug is not fixed yet even in VS2017.

PR #251, opened 2018-10-14
User friendly random number generator

masrourmouad:
Math.random(12) output a number between 0 and 12

### DIFF
--- a/src/optionals/gravity_math.c
+++ b/src/optionals/gravity_math.c
@@ -614,7 +614,6 @@ static bool math_pow (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uin
     RETURN_VALUE(VALUE_FROM_UNDEFINED, rindex);
 }
 
-// returns a random number between 0 and value
 static bool math_random (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
     #pragma unused(vm, nargs)
     gravity_value_t value = GET_VALUE(1);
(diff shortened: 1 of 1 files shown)


masrourmouad:
ok , working on it

masrourmouad:
Changes are made

marcobambini:
Thanks @masrourmouad, @hallzy is right, documentation needs to be updated to reflect the change.
Even if I am not sure about the value that your changes add.
What other people think?

masrourmouad:
Thanks @marcobambini for you comment, on my side I intend to render the function math.random
able to detect if the input is an integer or  a float and generate a number between 0 and input

marcobambini:
If we change the Math.rand method then, to be really useful, I think it should work in the following way:

- if no parameters are passed then the output will be a number between 0 and 1 (like the current version)
- if one parameter is passed it must be Int or Float and a number between 0 and the parameter will be returned
- if two parameters are passed they must be both Int or Float and a number between parameter1 and parameter2 will be returned
- any other parameter will be ignored

What do you think?

masrourmouad:
Nice idea

marcobambini:
It turns out that the problem is much more complicated and that rand() is a terrible API.
We need a much better multiplatform function (to support crypto and games) but unfortunately arc4random is not available on all platforms.

masrourmouad:
true, Wren have a good API , we need  a seed method .

marcobambini:
Fixed by 
https://github.com/marcobambini/gravity/commit/855d44a244ce8cf564d7609b34ee3ceb758e984a

PR #269, opened 2019-02-07
Added ENV class, switched C standard, minimal fix to CLI and debug helper.

IngwiePhoenix:
The most important changes in here are:

- `src/optional/gravity_env.{c|h}`: Implements a small class to represent the environment.
  * It currently lacks construction prevention.
  * It currently does not support iteration.
  * It does support map-like access (`ENV["PATH"]`).
  * It does support map-like acces with value setting (`ENV["foo"] = "bar"`).
  * It does perform type-checking (always expects a string).
  * It does return all available environment keys - although I feel as if this implementation is rather wobbly...
* `CMakeLists.txt`: After looking through the source for a long, long time, I realized that it is actually not using any of the C11 standard. Therefore, I adjusted it to what it is most conform to.

The small fix on the CLI to document the`-t` switch or the little debug helper which will ultimatively only forward it's call to `printf()` are minimal.

Anything you'd like me to change? :)


IngwiePhoenix:
This article: http://www.qnx.com/developers/docs/7.0.0/#com.qnx.doc.neutrino.lib_ref/topic/s/strlcat.html

Suggested that adding the `strlcpy` and `strlcat` functions would only require `libc`. Obviously, this is not true. I later learned that those are functions which originated from BSD - which immediately explains why this works perfectly on OS X but not on Linux.

Now, in the case of constructing a key-value pair into a `char` array, I actually think that I may even be able to use `strncpy`... however - I am not aware if this one is _really_ in `libc` already, or not. Or even `glib`.

Therefore I will just revert the `-lbsd` changes and fix the fact that I didn't `free()` in `gravity_env_set`, and wait for your thought on this. :)

IngwiePhoenix:
> malloc/free should be replaced by mem_alloc/mem_free

Done.

marcobambini:
I made several changes and I simplified your code, avoiding a lot of temp allocations.

IngwiePhoenix:
I see you changed `putenv` to `setenv`. Well unfortunately, this one is not available on Windows, while `putenv` is. :)

See here: https://stackoverflow.com/a/23073039/2423150

IngwiePhoenix:
https://github.com/marcobambini/gravity/commit/87ccab0e8ee0eb50df85d605343602815161d467#diff-d54ca8a7622e239e7a4d7ca4d0aef8d4R60

Not true, actually. It is used in the debug macro. Or does this not matter?

marcobambini:
_putenv seems available on Windows https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/putenv?view=vs-2017

IngwiePhoenix:
I couldn't find any clear explanation of wether to use `setenv` or `putenv` - so I decided to open a SO question: https://stackoverflow.com/questions/54605663/windows-setenv-vs-putenv-which-one-to-use-for-cross-platform-environment

marcobambini:
Put is better because you do not need to allocate any temp memory

marcobambini:
Windows should be fixed now.

IngwiePhoenix:
Looks fine to me. I dont have a Windows maschine at my hand right now, but from what I can gather, this should work fine. :)

PR #278, opened 2019-05-13
Cast pointer to uintptr_t instead of unsigned long.

iSLC:
Because long could be 32bit on other platforms (_ie. Windows_) regardless of architecture.

Although I don't quite understand that cast or it's purpose (_didn't look much into it either_). It looks quite awkward at a glance. Perhaps I should've looked deeper into the implementation.

Which means if that was the intended behavior. Then I apologize and please ignore this PR.


marcobambini:
Hi @iSLC thanks a lot for all your fixes. This is the only patch that I cannot accept because the JSON code was taken from https://github.com/udp/json-parser and all fixes/enhancements should be added directly to that repo.

Issue #280, opened 2019-06-17
Http support


jamesalbert:
I'm not sure if http support is better implemented built-in or through a library, but it's a generally useful feature.

I took a stab at handling get requests (super ugly, just got an mvp going) [shown here](https://github.com/jamesalbert/gravity/commit/2695e5b24bd663e0957598c21472006d19ab737f). Are there already plans to implement something like this? If not, with some feedback, is this something that can be added to the project?

IngwiePhoenix:
There was someone planning to work on passing features from a single-header socket library to Gravity.

Fortunately, making a wrapper for a library towards Gravity is not that difficult - but it does involve reading through the various headers, since not all functions are labeled very well (take the `marry_*` functions and macros for instance, which do array handling).

But for that, you do get a very low-level access to Gravity and values, reducing the layers of abstractions quite some and making access to the wrapper itself quite fast.

I was looking into Swig, but trying to write a wrapper generator using Swig is all but easy unfortunately...

jamesalbert:
@IngwiePhoenix is there a wrapper on github I can use for reference? I tried looking, but gravitational spams the search + it seems there are still open issues for module support

IngwiePhoenix:
https://github.com/marcobambini/gravity/blob/master/src/optionals/gravity_env.c
https://github.com/marcobambini/gravity/blob/master/src/optionals/gravity_env.h

This is one that I contributed. In general, you can find some good examples within the `optionals` folder.

jamesalbert:
Oh, I saw those. That's what I modeled my changes off of

IngwiePhoenix:
And yes - module support is in fact still missing. But with module support, people mean something like this:

```
import x from "y";
var x = require("y");
```

But you can still add functions to Gravity just fine.

jamesalbert:
think my link was poorly placed, fix it

PR #281, opened 2019-06-18
Http support

jamesalbert:
Just want to say I've had a lot of fun with gravity 😄

I created an issue #280 about adding http support. I'm still unsure if this is wanted at this point in the project. With feedback, I'd be happy to refactor this janky pr a bit and implement the rest for an mvp. Or you can tell me my code sucks and to go home.

I've been testing using a basic flask server and the following gravity script:
```
func main() {
    var get_tests = [
        [
            "host": "https://reddit.com"
        ],
        [
            "host": "https://api.github.com",
            "path": "/users/jamesalbert"
        ],
        [
            "host": "https://httpbin.org",
            "path": "/get"
        ]
    ]

    get_tests.loop(func (request) {
        var resp = Http.get(request)
        System.print("=== " + resp.Hostname)
        System.print(resp)
    })

    var post_tests = [
        [
            "host": "https://httpbin.org/post",
            "path": "/post",
            "data": [
                "pet": "cat"
            ]
        ]
    ]

    post_tests.loop(func (request) {
        var resp = Http.post(request)
        System.print("=== " + resp.Hostname)
        System.print(resp)
    })
}
```


src/shared/gravity_value.c
```diff
@@ -2252,12 +2351,23 @@ inline gravity_string_t *gravity_string_new (gravity_vm *vm, char *s, uint32_t l
     obj->s = (char *)s;
     obj->len = len;
     obj->alloc = (alloc) ? alloc : len;
+    obj->s = mem_alloc(NULL, obj->alloc);
```
jamesalbert: this was needed to concat, unsure of the consequences

src/shared/gravity_hash.c
```diff
@@ -174,6 +174,21 @@ gravity_hash_t *gravity_hash_create (uint32_t size, gravity_hash_compute_fn comp
     return hashtable;
 }
 
+gravity_hash_t *gravity_hash_create_from (gravity_hash_t *from_hash) {
```
marcobambini: `gravity_hash_dup` would probably be a better name

src/shared/gravity_value.c
```diff
@@ -2192,6 +2192,99 @@ void gravity_map_blacken (gravity_vm *vm, gravity_map_t *map) {
...
+        sprintf(substr, "%s\"%s\": %s", delimiter, key_string, value_map_string);
+        gravity_string_concat_cstring(NULL, map_string, substr);
+        return;
+    }
+
+    // should never reach this point
+    assert(0);
+}
+
+char *gravity_map_to_string(gravity_vm *vm, gravity_map_t *map) {
```
marcobambini: Why not to reuse `convert_map2string` defined in gravity_core.c?

src/shared/gravity_value.c
```diff
@@ -2192,6 +2192,99 @@ void gravity_map_blacken (gravity_vm *vm, gravity_map_t *map) {
...
+    gravity_string_t *map_string = gravity_string_new(NULL, "", 1000, 0);
+    gravity_hash_iterate(map->hash, gravity_map_stringify_iterator, map_string);
+    gravity_string_concat_cstring(NULL, map_string, "}");
+    return map_string->s;
+}
+
+gravity_value_t gravity_map_to_value (gravity_vm *vm, gravity_map_t *map) {
```
marcobambini: Not sure to understand the use case of this...

src/shared/gravity_value.c
```diff
@@ -2258,12 +2351,23 @@ inline gravity_string_t *gravity_string_new (gravity_vm *vm, char *s, uint32_t l
...
+    obj->s = mem_alloc(NULL, obj->alloc);
+    strcpy(obj->s, s);
     if (s && len) obj->hash = gravity_hash_compute_buffer((const char *)s, len);
 
     if (vm) gravity_vm_transfer(vm, (gravity_object_t*) obj);
     return obj;
 }
 
+inline void gravity_string_concat_cstring (gravity_vm *vm, gravity_string_t *obj, char *cstring) {
+    uint32_t len = (uint32_t)(strlen(obj->s) + strlen(cstring));
```
marcobambini: obj->len would be much better than strlen(obj->s)

src/shared/gravity_value.c
```diff
@@ -2258,12 +2351,23 @@ inline gravity_string_t *gravity_string_new (gravity_vm *vm, char *s, uint32_t l
...
+    uint32_t alloc = MAXNUM(len+1, DEFAULT_MINSTRING_SIZE);
+    obj->s = mem_realloc(NULL, obj->s, alloc);
```
marcobambini: Instead of always perform a men_realloc you should first check if (obj->alloc - obj->len) is big enough

src/shared/gravity_value.c
```diff
@@ -2258,12 +2351,23 @@ inline gravity_string_t *gravity_string_new (gravity_vm *vm, char *s, uint32_t l
...
+    obj->len = len;
+    obj->alloc = alloc;
+    strcat(obj->s, cstring);
```
marcobambini: memcpy would be faster (you already have the right pointer by using (obj->s + obj->len)

marcobambini (changes requested):
Please check my reviews

marcobambini:
Thanks @jamesalbert for your work on Gravity!

jamesalbert:
Awesome @marcobambini I'll try and tackle these after work. Np, it's a nice little language, I like the style

marcobambini:
Thanks @jamesalbert and please note that a unit test would be recommended.

src/shared/gravity_value.c
jamesalbert: @marcobambini Wouldn't you say there's a need for _to_value functions for all types? E.g. if I want to create a map with an key associated to an array, I'd have no way of doing that without doing something like:

```c
gravity_map_t *response = gravity_map_new(vm, 32);
gravity_list_t *headers = gravity_list_new(vm, 32);
gravity_map_insert(vm, response,
    gravity_string_to_value("Headers"),
    gravity_list_to_value(headers));
```

If there's a way for already doing something like this, let me know. I'm cleaning up the implementation at the moment, currently trying to get something like this going

jamesalbert: I use the map_to_value method in a similar fashion here: https://github.com/marcobambini/gravity/pull/281/files#diff-fb63d91a9c6432f884cfba6cc100f7d9R160

jamesalbert: Oh I guess VALUE_FROM_OBJECT might be helpful in this case 😎

jamesalbert:
@marcobambini I figured an http client without ssl support is pretty useless. I added openssl support and tried to make it as non-dependent as possible. If the user doesn't have openssl, they can set the optional make parameter `OPENSSL_ENABLED` to false. I only have my macbook (with openssl) to my disposal, so I added some docker to build and run the example (1 container with openssl, 1 without). All 3 environments are building for me, but I'm sure there's more we can do.

Let me know what you think.

marcobambini (approved):

marcobambini:
@jamesalbert I apologize for taking so long in reviewing your pull request but I made a lot of refactoring for optional classes. Can you please remove the files:
gravity_visualstudio/gravity.vcxproj
gravity_visualstudio/gravity.vcxproj.filters
so I can fix the remaining conflicts and merge your pull request?
Thanks a lot.

hggz:
@marcobambini not sure if this is still possible to be revisited. would it be possible to carry this work through with this pr's changes, if it currently isnt in the main release?

marcobambini:
Not planned

PR #283, opened 2019-07-24
Fix 22 missing dependencies in Makefile

vemakereporter:
Hi, I've fixed 22 missing dependencies reported.
Those issues can cause incorrect results when gravity is incrementally built.
For example, any changes in "src/compiler/gravity_ast.h" will not cause "src/compiler/gravity_ast.o" to be rebuilt, which is incorrect.
I've tested it on my computer, the fixed version worked as expected.
Looking forward to your confirmation.

Thanks
Vemake


marcobambini:
Hi @vemakereporter and thanks for your contribution.
Unfortunately I cannot accept it because the true power of this Makefile is its ability to automatically scans each .c file and rebuild the executable according. No need to keep it manually updated each time a new file is added to the build. List all files by enumeration is not an option here.

vemakereporter:
Hi @marcobambini , Thank you for your reply. Can you help us to confirm that the 22 missing dependencies are valid? If this is the case, we can try to figure out how to fix those issues without enumerating all the files. 

Thanks.

Issue #295, opened 2020-06-10
how to build it on windows with gcc(MinGW) and get .a file ?


acnologia000:
how can i build gravity on windows to get .a and .h files to embed in other tools ? 
also i would like for more detailed examples for embedding api along with compilation instructions

marcobambini:
Thanks to @foobit you can now use CMake to build static libraries on Windows.
If you are not familiar with CMake there is also a GUI helper available from https://cmake.org

PR #299, opened 2020-06-17
Updated CMakeLists.txt to support multiple targets

foobit:
Updated CMakeLists.txt to support the following targets:
- **gravity** cli
- **gravityapi** shared lib (.dylib, .dll, .so)
- **gravityapi_s** static lib (.lib, .a)

Tested:
- Windows (VC2019, MinGW64)
- Ubuntu (gcc)
- MacOSX (xcode)

#295 - builds with MinGW
`md build && cd build`
`cmake -G "MinGW Makefiles" ..`
`cmake --build .`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,7 @@
 cmake_minimum_required(VERSION 3.0)
+project(gravity VERSION 1.0 LANGUAGES C)
+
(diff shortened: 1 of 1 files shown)


marcobambini:
Thank you @foobit!

PR #301, opened 2020-06-18
Removed public include dependency on gravity_visualstudio/unistd.h

foobit:
Embedding Gravity on Windows (MSVC) platforms requires an extra public include dependency that exists _outside_ the Gravity source tree.

This pull request removes `gravity_visualstudio/unistd.h` and adds `shared/gravity_config.h`.  This will allow public include paths match on all platforms.

Additionally `gravity_config.h` can be used for future platform specific includes and configurations.


marcobambini:
I really like this change, thanks @foobit!

Issue #300, opened 2020-06-18
Public macros with C++ compatibility


foobit:
Gravity integration with C++ projects work great except for a few **VALUE_FROM** macros.

The following public facing macros use _designated initializers_ which is not supported in C++ (_unless you use C++20_):
- `VALUE_FROM_ERROR`
- `VALUE_FROM_OBJECT`
- `VALUE_FROM_INT`
- `VALUE_FROM_FLOAT`
- `VALUE_FROM_NULL`
- `VALUE_FROM_UNDEFINED`
- `VALUE_FROM_BOOL`
- `STATICVALUE_FROM_STRING`

I looked into doing a pull request using the same pattern as `gravity_string_to_value` to hide the inline initializers in c code.  However, this will require additional code changes in places where VALUE_FROM requires an _l-value_.

e.g. gravity_core.c `range_loop` uses `&VALUE_FROM_INT(...)`

What are your thoughts on this?

marcobambini:
I didn't know that this issue exists with C++.
I tried the following C++ example:

```c++
#include <iostream>
#include "gravity_value.h"
#include "gravity_macros.h"

int main(int argc, const char * argv[]) {
    gravity_value_t value = VALUE_FROM_NULL;
    
    return 0;
}
```

And I obtained a linker error: `Undefined symbol _gravity_class_null`.
Is this error related to the issue you are reporting?

foobit:
## Test project
https://github.com/foobit/cppdi


## C++ designated initializers
Looks like designated initializers support for C++ depends on platform, C++ dialect and compiler version.

## Windows 10
- Building for: Visual Studio 16 2019
- The C compiler identification is MSVC 19.26.28806.0

**C++11**
> set (CMAKE_CXX_STANDARD 11)
```
..\src\main.cpp(6): error C7555: use of designated initializers requires at least '/std:c++latest'
..\src\main.cpp(6): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax
```

_**Note:** even set to C++20 MSVC seems to not fully support it._
> set (CMAKE_CXX_STANDARD 20)
```
..\src\main.cpp(6): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax
```
---
MinGW
- The C compiler identification is GNU 8.1.0
- The CXX compiler identification is GNU 8.1.0

_**gcc 8+** support C+2a features_
```
succeeds
```

---
## Ubuntu 18.04
[...]

PR #303, opened 2020-06-20
GRAVITY_USE_HIDDEN_INITIALIZERS to hide public facing designated initializers when using C++

foobit:
Added `GRAVITY_USE_HIDDEN_INITIALIZERS ` define in gravity_config.h to solve the issue of using Gravity in pre C++20 projects.

Added `gravity_value_from_XXX` functions that implement the `VALUE_FROM_XXX` macros

Reference: #300 Public macros with C++ compatibility

### DIFF
--- a/src/shared/gravity_config.h
+++ b/src/shared/gravity_config.h
@@ -35,4 +35,12 @@ typedef int         mode_t;
 #include <unistd.h>
 #endif
 
+// check if the compiler supports designated initializers when using c++
+#ifndef GRAVITY_USE_HIDDEN_INITIALIZERS
+    #if __cplusplus && (__cplusplus < 202002L)
+        // C++ di requires C++20
+        #define GRAVITY_USE_HIDDEN_INITIALIZERS 1
+    #endif // __cplusplus
+#endif // GRAVITY_USE_HIDDEN_INITIALIZERS
+
 #endif // __GRAVITY_CONFIG__
--- a/src/shared/gravity_macros.h
+++ b/src/shared/gravity_macros.h
@@ -9,6 +9,8 @@
 #ifndef __GRAVITY_MACROS__
 #define __GRAVITY_MACROS__
 
+#include "gravity_config.h"
+
 #define AUTOLENGTH                          UINT32_MAX
 
 // MARK: -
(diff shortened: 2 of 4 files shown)


marcobambini:
Thanks @foobit your code is pretty good.
I am going to merge your pull request right now.
Just to be extra sure I only ask you to run the unit test suite on Windows when you have some minutes: https://marcobambini.github.io/gravity/#/quickstart?id=unit-tests

foobit:
#304 -  I had to do a pull request to fix running unit tests on Windows. 

With latest on master  `gravity -t test\unittest`:
-  **MinGW** passes all test
- **MSVC** fails two tests `int_bit.gravity` and `radian_degree_conversion.gravity`

_Note:_ MSVC also fails those unit tests pre #303 merge.

marcobambini:
Thank you very much @foobit, the Windows version really needed some care.

marcobambini:
Both `int_bit.gravity` and `radian_degree_conversion.gravity` contains maths operations, I am wondering why both fails on Windows.

foobit:
```
Test file: int_bit.gravity
Test path: /sdks/gravity/test/unittest/int_bit.gravity
Int bitwise operations.
RUNTIME ERROR: Unable to perform operator LSHIFT on object
FAILURE
```
```swift
var n = 999999982873;  // prime number
n <<= 2;               // prime number
```
---

```
Test file: radian_degree_conversion.gravity
Test path: D:\sdks\gravity\test\unittest/radian_degree_conversion.gravity
Test Int.radians, Int.degrees
FAILURE
```
I didn't debug this yet. Just looking at the Gravity source, could be a rounding error ¯\\\_(ツ)_/¯

marcobambini:
I found the issue for the test `int_bit.gravity`.

It happens on Windows only because Visual Studio didn't support computed goto (not sure if it is now supported by recent versions) and so all the VM loop is being interpreted as a giant `switch` statement.

```c
// gravity_value.c:93

#ifdef _MSC_VER
#undef GRAVITY_COMPUTED_GOTO
#define GRAVITY_COMPUTED_GOTO               0           // MSVC does not support it
#endif
```

The issue is due to the macro `CHECK_FAST_BINARY_BIT(r1, r2, r3, v2, v3, <<);` used in gravity_vm.c:796 which at the end is expanded to:

```c
do {SETVALUE(_r1, VALUE_FROM_INT(_v2 _OP _v3)); DISPATCH();} while(0)
```

The problem is the DISPATCH macro that in case of GRAVITY_COMPUTED_GOTO equals to 0 it is expanded to a `break;` statement. The `break;` is then executed inside a do/while and it simply exists from the inner loop instead of exit from the main `switch/case` loop.

I am reclutant to remove the do/while (0) trick because it can avoid a lot of errors with macro usage, especially in if statements.

I am not sure if a solution exists without removing the do/while (0) trick. Any idea?

marcobambini:
@foobit fixed by introducing DISPATCH_INNER macro in https://github.com/marcobambini/gravity/commit/b51227de5524b001b236f0552ef3e32f3fc0b043

marcobambini:
@foobit to be able to compile with Visual Studio 2019 I had to remove some inline declarations from the c files. Have you had the same issue? Do you have an explanation or a solution?

foobit:
The `DISPATCH_INNER` fixed the unit tests for VS2019.

I noticed the issues with the inline declarations. I inspected the .vcproj directly and someone from the original vcproj added the fix. I added the same setting for the CMakeLists.txt file when I updated it.

```cmake
# MSVC does not like private static inlining
list(APPEND GRAVITY_PRIVATE_DEFINITIONS "inline=")
```
I assume if those static inline were in public headers, it won't optimize out. I haven't done a test on that theory.

marcobambini:
What do you think if I removed entirely the inline declarations? I don't think that modern compilers use it in any way.

foobit:
I'm more on the stance that compilers are now pretty smart these days. Most will optimize quite nicely without the inline hints.

I'm sure you won't notice a difference by removing them.

PR #306, opened 2020-06-26
added example on inline string formatting

foobit:
I'm not familiar with swift. However, I discovered the string interpolation in the unit tests.  Added a brief comment about it in the docs.

*Related:* It looks like Gravity scripts and strings are **UTF8**.  I was able to verify when reading a Gravity string from C.

Should there be a comment somewhere indicating they are UTF8?


marcobambini:
You are right @foobit, I added a note about UTF-8 support with https://github.com/marcobambini/gravity/commit/b15485b23bcfde962478c1ce570716f379063d1c

PR #310, opened 2020-07-02
fixed CMake Win32 shared library export

foobit:
CMakeLists.txt did not define `BUILD_GRAVITY_API` for the shared library.  This caused exporting to not work on Win32.


marcobambini:
Thanks!

PR #323, opened 2020-10-11
fix some potential null pointer dereference bugs.

Yunlongs:
When we try to alloc some memories and the size is directed by some variables or parameters, we should better to check ther returned pointer whether is NULL.
Because if the alloc function failed, it will return a NULL pointer. When we try to read or write the NULL pointer, it will cause `Segmentation fault (core dumped)`.  It does not provide enough information to tell what cause this crash.

We should better to check if these pointer is null.  These will make our project more robust.


marcobambini:
My assumption is that on modern Operating Systems malloc never fail. I could be wrong if Gravity is used in embedded or older systems.

I cannot accept this patch because I would not use assert at runtime and I would need to better handle all memory allocations. Thanks for pointing out the issue, I'll invest some time for a more general solution.

PR #327, opened 2020-10-26
fix size_t comparison of negative in file_read()

rwhitworth:
size_t is unsigned, so comparison against -1 effectively causes undefined behavior

Fixes:
```
src/utils/gravity_utils.c:150:19: runtime error: implicit conversion from type 'int' of value -1 (32-bit, signed) to type 'unsigned long' changed the value to 18446744073709551615 (64-bit, unsigned)
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/utils/gravity_utils.c:150:19 in
```


marcobambini:
Thanks!

PR #328, opened 2020-10-26
fix signed char to unsigned char conversion in utf8_charbytes()

rwhitworth:
Marco, you probably recall that I'm not a strong C programmer, so take this fix with a grain of salt. I used https://wiki.sei.cmu.edu/confluence/display/c/STR34-C.+Cast+characters+to+unsigned+char+before+converting+to+larger+integer+sizes as my guide for suggesting a fix for this one. Please let me know if this is not appropriate or does not solve the problem in the right way.

Fixes:
```
src/utils/gravity_utils.c:473:23: runtime error: implicit conversion from type 'char' of value -27 (8-bit, signed) to type 'unsigned char' changed the value to 229 (8-bit, unsigned)
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/utils/gravity_utils.c:473:23 in
```

### DIFF
--- a/src/utils/gravity_utils.c
+++ b/src/utils/gravity_utils.c
@@ -469,7 +469,7 @@ end_repl_str:
  */
 
 inline uint32_t utf8_charbytes (const char *s, uint32_t i) {
-    unsigned char c = s[i];
+    unsigned char c = (unsigned char)s[i];
     
     // determine bytes needed for character, based on RFC 3629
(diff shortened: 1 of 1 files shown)


marcobambini:
In this case c is already an unsigned char so just the explicit cast is missing.
UBSan is wrong here (even if it is formally correct).
I accept this patch because it is better to be pedantic that to omit some important details.
Thanks.

PR #330, opened 2020-10-26
fix uint32_t to char conversions in utf8_encode()

rwhitworth:
This resolves the runtime errors, but I'm quite unsure if this is the proper way to resolve the issue. I compared to the Wren source, only to notice the variable types have been changed, which may result in this type mismatch?

Input file:
[utf8_encode_char.txt](https://github.com/marcobambini/gravity/files/5436642/utf8_encode_char.txt)

Fixes:
```
src/utils/gravity_utils.c:518:18: runtime error: implicit conversion from type 'unsigned int' of value 170 (32-bit, unsigned) to type 'char' changed the value to -86 (8-bit, signed)
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/utils/gravity_utils.c:518:18 in
```

### DIFF
--- a/src/utils/gravity_utils.c
+++ b/src/utils/gravity_utils.c
@@ -496,37 +496,37 @@ uint32_t utf8_encode(char *buffer, uint32_t value) {
     
     if (value <= 0x7f) {
         // single byte (i.e. fits in ASCII).
-        *bytes = value & 0x7f;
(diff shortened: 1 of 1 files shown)


marcobambini:
See my previous comment for https://github.com/marcobambini/gravity/pull/328
Thanks.

PR #332, opened 2020-11-08
Added class inheritance documentation

RobLoach:
Found that documentation surrounding class inheritance was missing. So, here you go!


marcobambini:
Thank you so much @RobLoach I really appreciate your help!

PR #334, opened 2020-11-08
cmake: Split up the CLI and Library CMake definitions

RobLoach:
Having the CLI CMake file and Libraries CMake file in different directories allows third-party applications to reference the library file directly, without having to compile the executable...

``` cmake
add_subdirectory(vendor/gravity/src)
target_link_libraries(myproject gravityapi)
```

This is the recommended pattern for [Modern CMake](https://cliutils.gitlab.io/modern-cmake/chapters/basics/structure.html).

@foobit Mind a review? It's working for me. Can reference just the source directory, and still build the executable with cmake no problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,10 +3,14 @@ project(gravity VERSION 1.0 LANGUAGES C)
 
 set(CMAKE_C_STANDARD 99)
 
+option(BUILD_CLI "Build the command line interface" ON)
+
 # ----------------------------------------------------------------
 # Library
 add_subdirectory(src)
 
 # ----------------------------------------------------------------
 # Command Line Interface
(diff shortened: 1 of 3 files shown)


foobit:
Your changes worked fine for me.  The split is nice. Tested cmake build with Visual Studio 2019 (Windows) and gcc 10 (Ubuntu). 

The only thing I would suggest is adding a cmake option to skip adding the cli subdirectory (have it default to on).  That would be useful for users who just embed the library.

The _Modern CMake_ also suggested to not GLOB or AUX_SOURCE. But that can be for another issue/pull.

marcobambini:
I am not a cmake expert but it seems a good change.
Thanks @RobLoach and @foobit for your help!

PR #335, opened 2020-11-14
Remove readline() from gravity utils

RobLoach:
This change removes `readline()` from `gravity_utils` because `getline()` isn't available on some platforms, causing fatal errors during compilation of `gravityapi`. Since REPL isn't implemented yet, `readline()` isn't used anywhere anyway.

Would be optimal to implement it as part of `gravity.c` when we get REPL in.


marcobambini:
Seems very reasonable to me.
Thanks @RobLoach.

Issue #21, opened 2017-03-07
Input from stdin?


jkleiser:
Does the System class have methods for reading from stdin?
I can see no mention of it here: https://marcobambini.github.io/gravity/system.html

marcobambini:
Not yet, I'll probably add it as soon as I'll setup a REPL mode.

marcobambini:
Fixed by https://github.com/marcobambini/gravity/commit/f67c05b18c14778ee17853430dfe6774dc98e33b

PR #342, opened 2021-01-02
Add System.input() for input from stdin

bynect:
Solves issue #21 adding a simple method for getting input from the user, even without a REPL.
system_input uses fgets instead of scanf and should be safe for buffer overflows.
The buffer size for now is 1024 because this seemed to be the most used buffer size in the repo.

### DIFF
--- a/src/runtime/gravity_core.c
+++ b/src/runtime/gravity_core.c
@@ -3141,6 +3141,15 @@ static bool system_print (gravity_vm *vm, gravity_value_t *args, uint16_t nargs,
     return system_realprint(vm, args, nargs, rindex, true);
 }
 
+static bool system_input (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+    #pragma unused(args,nargs)
+    char buffer[1024];
+    fgets(buffer, 1024, stdin);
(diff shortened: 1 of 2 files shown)


marcobambini:
Thanks a lot @bynect!

src/runtime/gravity_core.c
```diff
@@ -3141,6 +3141,15 @@ static bool system_print (gravity_vm *vm, gravity_value_t *args, uint16_t nargs,
     return system_realprint(vm, args, nargs, rindex, true);
 }
 
+static bool system_input (gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+    #pragma unused(args,nargs)
+    char buffer[1024];
+    fgets(buffer, 1024, stdin);
+    // remove trailing newline captured by fgets
+    buffer[strlen(buffer) - 1] = 0;
```
bynect, line 3149: Note that here the newline if removed from the buffer, so remove this line of code if the terminator is needed.

marcobambini: I improved that function with https://github.com/marcobambini/gravity/commit/3161cde05ac1735c9f6f7f5322373530c169337f

Issue #235, opened 2018-07-03
[feature request] Command line arguments


michal037:
Hello!

What do you think to add command line arguments to the interpreter?

```gravity
./gravity myfile.gravity arg1 arg2 arg3
```
And perhaps a class or variable for read it.

windyflying:
It's cool

IngwiePhoenix:
The CLI program may simply pass `argv`/`argc` to the main function. Take this concept code for example:

```c
gravity_list_t *list = gravity_list_new(vm, argc);
for(int i=0; i<argc; i++) marray_push(gravity_value_t, VALUE_FROM_STRING(argv[i]), list->array);

uint32_t nparams = 1;
gravity_value_t params[] = { VALUE_FROM_LIST(list) };

gravity_value_t main_fn = gravity_vm_getvalue(vm, "main", strlen("main"));
gravity_vm_runclosure(vm, VALUE_AS_CLOSURE(main_fn), NULL, params, nparams);

gravity_value_t res = gravity_vm_result(vm);
// ...
```

This would allow this signature:

```swift
func main(args) {
  System.print(args);
}
```

Where `args` is now the original `argv`/`argc` translated into a `List` object.

IngwiePhoenix:
A convenience macro may be used to simplify this process - i.e.: `VALUE_FROM_STRINGARRAY(_vm, _array, _len)`. It would essentially do what the first two lines of the above listing do and return a `gravity_value_t` immediately.

marcobambini:
Fixed by https://github.com/marcobambini/gravity/commit/11efd1596a602c309339e902e6a16f080b885b2f

Issue #113, opened 2017-03-31
Ability to pass arguments to a gravity script


hallzy:
for instance:

```bash
gravity add.gravity 2 4 5
```

Should return "11" assuming the code is something like:

```swift
func main() {
  var sum = 0
  _args.loop(func (value) {
    sum += value
  })
  return sum
}
```

Basically, just a way to get values from the command line would be quite useful

marcobambini:
Feature request moved to low priority wiki page: https://github.com/marcobambini/gravity/wiki/Low-priority-features

marcobambini:
Fixed by https://github.com/marcobambini/gravity/commit/11efd1596a602c309339e902e6a16f080b885b2f

PR #343, opened 2021-01-03
Add a way to pass command line arguments

bynect:
This pr addresses #113 and #235 adding a way to pass arguments to the script. Unlike the approach suggested by IngwiePhoenix in #235 here the arguments are not passed to the main function but are registered as properties of the ENV class. I am unsure if the System class is more appropriate than the ENV class, but this should be chosen by the maintainer/community.

Pros of this approach:
- Arguments are accessible from anywhere in the program and not only in the main function
- Arguments are not passed directly to the vm / run function
- Passing CLI arguments is optional and require a separate register function 

Cons:
- Arguments are tight to an optional class
- To register the arguments you have to include the gravity_opt_env.h header

Possible improvements:
- Unify argc and argv to a single list called args, where argc is simply the length of the list containing argv
- Move maybe in the System class (?)

I also edited the documentation for the ENV class.
Feel free to modify this pr.

### DIFF
--- a/src/optionals/gravity_opt_env.c
+++ b/src/optionals/gravity_opt_env.c
@@ -27,6 +27,9 @@
 static gravity_class_t              *gravity_class_env = NULL;
 static uint32_t                     refcount = 0;
 
+static int                          argc = -1;
+static gravity_list_t               *argv = NULL;
+
 /**
  * Wraps `getenv()` to be used with Gravity.
  *
@@ -96,6 +99,24 @@ static bool gravity_env_keys(gravity_vm *vm, gravity_value_t *args, uint16_t npa
     RETURN_VALUE(VALUE_FROM_OBJECT(keys), rindex);
 }
 
+static bool gravity_env_argv(gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+    #pragma unused(vm, args, nargs)
+    
+    if (argv)
+        RETURN_VALUE(VALUE_FROM_OBJECT(argv), rindex);
+
+    RETURN_VALUE(VALUE_FROM_NULL, rindex);
+}
+
+static bool gravity_env_argc(gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+    #pragma unused(vm, args, nargs)
+
+    if (argc != -1)
+        RETURN_VALUE(VALUE_FROM_INT(argc), rindex);
+
+    RETURN_VALUE(VALUE_FROM_NULL, rindex);
+}
+
 // MARK: - Internals -
 
 static void create_optional_class (void) {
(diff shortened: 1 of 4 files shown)


src/optionals/gravity_opt_env.c
```diff
@@ -124,14 +152,26 @@ void gravity_env_register(gravity_vm *vm) {
     gravity_vm_setvalue(vm, GRAVITY_CLASS_ENV_NAME, VALUE_FROM_OBJECT(gravity_class_env));
 }
 
+void gravity_env_register_args(gravity_vm *vm, uint32_t _argc, const char **_argv) {
+    gravity_value_t tmp[_argc];
+    for (int i = 0; i < _argc; ++i) {
+        tmp[i] = VALUE_FROM_CSTRING(NULL, _argv[i]);
```
bynect, line 159: This feel a little hack-ish, is there a better way to append an item to a gravity_list_t?

marcobambini: It should be rewritten as:
```swift
void gravity_env_register_args(gravity_vm *vm, uint32_t _argc, const char **_argv) {
    argc = _argc;
    argv = gravity_list_new(vm, argc);
    for (int i = 0; i < _argc; ++i) {
        gravity_value_t = VALUE_FROM_CSTRING(vm, _argv[i]);
        marray_push(gravity_value_t, argv, value);
    }
}
```

src/optionals/gravity_opt_env.c
```diff
@@ -96,6 +99,24 @@ static bool gravity_env_keys(gravity_vm *vm, gravity_value_t *args, uint16_t npa
...
 }
 
+static bool gravity_env_argv(gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+    #pragma unused(vm, args, nargs)
+    
+    if (argv)
+        RETURN_VALUE(VALUE_FROM_OBJECT(argv), rindex);
+
+    RETURN_VALUE(VALUE_FROM_NULL, rindex);
+}
```
marcobambini, line 109: What about:
```swift
static bool gravity_env_argv(gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
    #pragma unused(args, nargs)
   RETURN_VALUE((argv) ? VALUE_FROM_OBJECT(argv) : VALUE_FROM_NULL, rindex);
}
```

src/optionals/gravity_opt_env.c
```diff
@@ -96,6 +99,24 @@ static bool gravity_env_keys(gravity_vm *vm, gravity_value_t *args, uint16_t npa
...
+
+static bool gravity_env_argc(gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
+    #pragma unused(vm, args, nargs)
+
+    if (argc != -1)
+        RETURN_VALUE(VALUE_FROM_INT(argc), rindex);
+
+    RETURN_VALUE(VALUE_FROM_NULL, rindex);
+}
+
```
marcobambini, line 119: What about:
```swift
static bool gravity_env_argc(gravity_vm *vm, gravity_value_t *args, uint16_t nargs, uint32_t rindex) {
    #pragma unused(args, nargs)
    RETURN_VALUE((argc != -1) ? VALUE_FROM_INT(argc) : VALUE_FROM_NULL, rindex);
}
```

bynect: Seems better, today I'll commit those fixes.

marcobambini:
Thanks a lot @bynect

PR #354, opened 2021-05-08
Clang: Fixed implicit conversion from long to double

SheatNoisette:
This simple pull request fixes an implicit conversion warning on Clang 12.0 :
```
src/optionals/gravity_opt_math.c:908:45: warning: implicit conversion from 'long' to 'double' changes value from 9223372036854775807 to 9223372036854775808 [-Wimplicit-const-int-float-conversion]
 ```


marcobambini:
Thanks!

PR #365, opened 2021-06-15
fix read size different from fs-info file-size due to line-ending

tDwtp:
fix line ending causing fsize != fsize2 in gravity_utils.d/file_read() due to not reading in binary mode.

On Windows when opening a file in text mode `\r` is ignored (and not read) after each `\n`. This causes files never to be loaded and giving an uninformative error just being: "`Error reading file [...]`". `\r`s are ignored by the compiler anyway.


marcobambini:
Thanks a lot @tDwtp!

Issue #368, opened 2021-06-17
put `gravity_hash_*` from `shared/gravity_value` inside `hared/gravity_hash`


tDwtp:
I don't see why they should be there.
Pull request in a minute...

marcobambini:
Just accepted, thanks.

marcobambini:
P.S. there is a typo in `gravity_hash_finteralfree`, it should be `gravity_hash_interalfree `

PR #369, opened 2021-06-18
move hash functionality

tDwtp:
see #368 for more details.
In short: hash functions are expected to be in `g*_hash.h` When dynamically linking this makes more sense.


marcobambini:
Thanks @tDwtp

PR #370, opened 2021-06-18
Dynamic linked cli

tDwtp:
Dynamically link cli, so expanding by a native library is easier. 
Linking against an EXE is annoying to do. Static linking will introduce versioning problems, when the native-library was compiled from another version. nasty (security-related) bugs can occur. Neither are "good" solutions.

Explanation of `gravity_common.h`: there are multiple files using `GRAVITY_API` instead of defining it in multiple files or loading all of `gravity_value.h` i centralized it to this one header. If something is used header wide, it should be placed here
Also i might change `gravity_config.h` later down the line to have a better overview of the configs similar to Lua.

I expose as much API as possible in order to be more attractive to developers. Especially as most are used by Gravity itself.
Exposing headers in a seperate directory (`include`) could be helpful.


tDwtp:
to much exposure. Not wanted by Owner.

PR #374, opened 2021-06-21
Rework cli

tDwtp:
Please note: I had to export (or hide) certain symbols:
- `gravity_json.h` (required) these are used mutliple times inside different exposed functions, some even beeing essential.
- `gravity_utils.h` (optional and handy) these are really handy for expanding gravity and are simple to understand.
- `gravity_debug.h` (optional) for (custom/gravity-specific) editors these symbols will come in handy ~~
- `gravity_compiler.h#gravity_compiler_ast` I dont think you wanted me to export *that* much, this would require `gravity_token.h` and `gravity_node.h` to be (maybe) *mostly* exposed to make any sense of it. Subsequently there would be more headers. I chose to hide it, as it is not used anywhere.

Regarding `gravity_vm_register_args`: I would expose optionals only for the dll (please evaluate this). but for the executable it would not make any sense. besides this is a little less dependent on the optionals. The call does not seem to do anything which requires this specific optional anyway? Maybe it should be moved to `gravity_vm.h`/`*.c`

I added Build-Targets and cleaned up the exports. I also want to add a `dyn` folder for the dynamically linkable libraries... currently it looks cluttered.

Also: we should export the `computed_property_*`/`convert_value2*` functions. They are really helpful and a headder should either have alle symbols exported or none.

*SOMETHING* needs to be bundled with the executable to make expansion possible. a `*.lib`/`lib*.a` is the absolute minimum. You have to die one death. headers should also be included. however... on windows a dllimport is needed. Thats why I created the `gravity_common.h` header. We can choose to put it there. (also it makes exported symbols not include gravity_value...)


marcobambini:
Not planned

PR #375, opened 2021-07-04
Build shared object on BSDs; remove unneeded WITH_GETLINE.

ibara:
Hello --

The BSDs are building with a -D_WITH_GETLINE switch that appears unreferenced in the code itself.
Additionally, there's no technical reason why not to build the shared library (if wanted) either, so make it available.

Thanks!


marcobambini:
Thanks @ibara!

PR #376, opened 2021-07-10
bzero is not cross-platform and deprecated.

iSLC:
The bzero() function is deprecated (marked as LEGACY in POSIX.1-2001); use memset(3) in new programs.


marcobambini:
Thanks @iSLC but there is already a macro that takes care of unsupported systems (in `gravity_config.h`):
```c
#if (!defined(HAVE_BZERO) || !defined(bzero))
#define bzero(b, len) memset((b), 0, (len))
#endif
```

iSLC:
I would like to add that it should probably be moved outside the `_MSC_VER` macro scope. Because there's more then MSC on windows. Like MinGW for example. Which it doesn't cover.

PR #377, opened 2021-07-21
Update Travis CI badge

pamarcos:
The current badge points to deprecated travis.org which redirects to travis.com, so we'd rather point to it directly :)


marcobambini:
Oh yes @pamarcos thanks!

pamarcos:
You are very welcome. Thanks for your work on Gravity @marcobambini, much appreciated

PR #378, opened 2021-07-21
Add unit test run to the CI

pamarcos:
Add unit test run to the CI to ensure there are no regressions introduced

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,6 @@ os:
 compiler:
   - gcc
   - clang
-script: make
+script:
+  - make
+  - test/unittest/run_all.sh
--- a/test/unittest/run_all.sh
+++ b/test/unittest/run_all.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+# This script is intended to be used either by the CI or by a developer that
+# wants to check that no unit tests have been broken to ensure there are no
+# regressions. At the end, it reports how many tests were successful. However,
+# using grep we can easily filter out everything except for those tests that
+# time out. e.g. ./run_all.sh | grep -i timeout -B1
+
+set -u -o pipefail
+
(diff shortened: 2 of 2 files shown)


marcobambini:
Thank you so much @pamarcos for including the unit testing in the CI.
I suspect that the `max_recursion.gravity` test should be tuned to be able to run in a virtual machine.

pamarcos:
You're welcome @marcobambini! I saw that the CI on master is failing on macOS because `max_recursion.gravity` is timing out.
On my Ubuntu 20.04 it gets killed by the OS instead because of OOM:

```
[42085.761977] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/user.slice/user-1000.slice/session-2.scope,task=gravity,pid=1320384,uid=1000
[42085.761988] Out of memory: Killed process 1320384 (gravity) total-vm:23073076kB, anon-rss:10525032kB, file-rss:36kB, shmem-rss:0kB, UID:1000 pgtables:20652kB oom_score_adj:0
[42086.077446] oom_reaper: reaped process 1320384 (gravity), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
```

I know the CI at the moment reflects current status, but we may consider either tweaking the recursion unit test to pass in macOS, or disable that test for macOS's CI :thinking:

marcobambini:
@pamarcos there was an issue in that test, I am working on a fix right now

marcobambini:
@pamarcos should be fixed now by https://github.com/marcobambini/gravity/commit/a135732f5b0cff4d194a3a4c54f7c02f1700a379

pamarcos:
> @pamarcos should be fixed now by a135732

Cool! Thanks a lot for your effort to fix this so quick, @marcobambini! :partying_face: 
The great news is that now that all of unit tests are passing, the CI time spent went from > 2mins to slightly less than 30 seconds. Excellent :+1:

PR #381, opened 2021-09-18
Fix unsigned/signed comparison warnings

SamuelMarks:
WiP

### DIFF
--- a/src/cli/gravity.c
+++ b/src/cli/gravity.c
@@ -449,7 +449,12 @@ int main (int argc, const char* argv[]) {
         if (type == OP_INLINE_RUN) {
             char *buffer = mem_alloc(NULL, size+1024);
             assert(buffer);
-            size = snprintf(buffer, size+1024, "func main() {%s};", input_file);
+            size =
+#if defined(_WIN32) || defined(__WIN32__) || defined(__WINDOWS__)
+                _snprintf_s(buffer, size+1024,  sizeof buffer, "func main() {%s};", input_file);
+#else
+                snprintf(buffer, size+1024, "func main() {%s};", input_file);
+#endif
             source_code = buffer;
         }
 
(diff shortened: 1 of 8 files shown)


marcobambini:
Thanks, and sorry this sat so long. It is marked WiP, and reading it again now it would introduce three regressions, so I would rather close it than leave it open indefinitely.

* **`gravity_semacheck2.c`** — `for (size_t i=len-1; i>=0; --i)` never terminates: `i >= 0` is always true for an unsigned type, so after `i == 0` it wraps to `SIZE_MAX` and `gnode_array_get` reads far out of bounds. `lookup_identifier` runs on essentially every identifier, so this is not an edge case.
* **`gravity_core.c`, `string_storeat`** — `register size_t index` makes the next line dead code: `if (index < 0) index = string->len + index;` can never fire, so negative indexing stops working. `var s = "hello"; s[-1] = "O"` returns `hellO` today and would become an out-of-bounds error.
* **`gravity.c`** — `_snprintf_s(buffer, size+1024, sizeof buffer, ...)`: `buffer` is a `char *`, so `sizeof buffer` is the pointer size rather than the buffer's. The count argument would be 8.

`gravity_env_register_args` moving from `uint32_t` to `size_t` also changes a public header, which breaks embedders in exchange for a warning.

The warnings themselves are worth cleaning up, and parts of this are correct on their own — `list_remove`'s `count`, the loop counters in `string_count`, the `(char)c` cast in `internal_file_iread`. If you want to take another pass with just those, it is genuinely welcome. But it needs to go file by file with the bounds checks read alongside each change: as this diff shows, a mechanical signed → unsigned swap changes what a comparison means, and the compiler stops warning precisely where the code stopped being right.

Closing. Thanks all the same.

SamuelMarks:
Yowza, ~5 years for a response then you close it immediately with very addressable concerns

PR #401, opened 2022-12-27
Fix Windows Line Number

### LINKED ISSUE #389: Line numbers in error messages are wrong
I introduced different syntax errors in a running code.
The resulting error messages are always larger than the line in question.
Subsequent detected errors did not conform to equally offset numbers.

How to debug a Gravity script?

mwasplund:
fixes #389

The lexer was incorrectly peeking at the current character when seeing if a carriage return also has a trailing line feed. Updated to use peek next to correctly skip over the LF.

Also added a script to run unit tests. These tests pass with and without my change, so I wonder if there is a way to verify the exact line information in the error code.


test/unittest/run_all.cmd
```diff
@@ -0,0 +1,38 @@
...
+@REM This script is intended to be used either by the CI or by a developer that
+@REM wants to check that no unit tests have been broken to ensure there are no
+@REM regressions. At the end, it reports how many tests were successful.
+
+set SCRIPT_DIR=%~dp0
+set GRAVITY_BIN=%SCRIPT_DIR%/../../gravity_visualstudio/bin/cli/x64/Debug/gravity.exe
+set /a tests_total=0
+set /a tests_success=0
+set /a tests_fail=0
+set /a tests_timeout=0
```
mwasplund, line 13: I am unsure what the best replacement is for timeout, so I removed the timout checks for now.

marcobambini:
Thanks — you found the right bug and the diagnosis is exact. Fixed in daefac8, built on your patch with two corrections.

Your reading is correct: `is_newline` used `PEEK_CURRENT` for the character *after* `c`, which only holds where the caller had already consumed `c` (`lexer_scan_comment`, `gravity_lexer_skip_line`). In `gravity_lexer_next`, `c` was still the character sitting at the current offset, so the CR never saw the LF next to it and the pair was counted as two line breaks. On a CRLF source an error on row 5 was reported on row 9.

Two of the call sites needed more than the extra arguments:

* **`lexer_scan_string`** — `is_newline` consumes with `NEXT`, which does not run `INC_TOKLEN`. Once `PEEK_NEXT`/`PEEK_NEXT2` are passed in, the CR of a CR+LF pair gets eaten there and only the LF reaches `next_utf8`, so the literal's byte count comes out one short per line: `"line1<CRLF>line2"` measures 11 instead of 12 and loses its last character. NEL and LS have the same problem — `is_newline` eats their lead bytes and `next_utf8` then chokes on a bare continuation byte. That site now consumes the whole terminator itself and counts every byte into the token.
* **`lexer_scan_number`** — nothing has to be consumed there, the terminator is not part of the number, so it only tests for one and leaves it to `gravity_lexer_next`.

`is_newline` was also split in two: `newline_length()` is a pure predicate returning the size of the terminator, and `is_newline()` is the consuming variant for the two callers that have already read `c`.

To your question at the end — yes, `error_row` and `error_col` in the `#unittest` block are exactly how to assert line information. There is now `test/unittest/bugfix_crlf_lineno.gravity`, stored with CR+LF endings (with a `.gitattributes` rule so `core.autocrlf` cannot rewrite it) and asserting both. It fails on the old lexer and passes on the new one.

Not applied: the `v142` → `v143` toolset bump in the `.vcxproj` files and `run_all.cmd`. The Visual Studio projects are not built by CI, and pinning a newer Visual Studio in the same change as a lexer fix is not something I want to do blind.

Closing since the fix is in. Thanks for the report and the patch.

PR #411, opened 2023-08-05
Add tcc support (and fix msbuild without visual studio install)

tDwtp:
This adds [tcc](https://bellard.org/tcc/) support. It should also fix some stuff with msbuild (commandline ithout visual studio).

The changes for make are of semantic nature.
Explanation:
Usually to distinguish between recipies and setup, you *have to use* tabs for recipies, but setup should be done with space indentation. When using spaces as indentation after a recipie it will not be interpreted as a call.
Also: some syntax-highlighter specifically require space indentation to distinguish between recipies and setup


marcobambini:
Thanks — one part of this turned out to be a real bug and is now in; the rest has aged out.

**Applied** in 40a26f2: `#ifdef WIN32` → `#ifdef _WIN32` in `gravity_utils.c`, `gravity_opt_file.c` and the CLI. This is worse than a tcc problem. `gravity_utils.h` already keys off `_WIN32` for `windows.h` and for `DIRREF`, while the `.c` files were keyed off `WIN32`, which no compiler defines on its own — the Visual Studio projects define it in exactly two configurations (`Debug|Win32` and `Release|Win32` of `gravity.vcxproj`) and nowhere else. So every x64 MSVC build was compiling the POSIX bodies — `opendir`/`readdir` — against a `DIRREF` that is a `HANDLE`. MinGW and tcc land in the same place. Good catch.

**Not applied:**

* the Makefile changes — it has been rewritten since (dependency tracking, per-target object lists, a `jsontest` target), so the diff no longer applies. Also `.PHONY: gravity$(EXEXT) example$(EXEXT)` marks real file targets as phony, which makes them rebuild unconditionally.
* `rm.bat` — a general-purpose `rm` shim at the repo root to serve one recipe is a lot of surface for the problem.
* the `WIN32_FIND_DATA` → `_WIN32_FIND_DATA` edits inside comments — those are Windows API type names and the rename makes the comments wrong.
* `$expr`/`$value` → `expr`/`value` in `gravity_codegen.c` — you are right in principle, `$` in an identifier is a GNU extension and not C99, but gcc, clang and MSVC all accept it and the `$` is deliberate: it marks the registers as compiler-generated in code that also emits Gravity source in its comments. Worth doing as its own change, with names that keep that distinction.

On tcc generally: CI builds Linux and macOS with gcc and clang, so tcc-specific configuration would be carried without anything checking it stays correct. I would rather not take that on. If tcc support matters to you, a CI job that builds it is the thing that would make it maintainable.

Closing — the fix that mattered is in. Thanks for finding it.

PR #427, opened 2024-07-08
Updated makefile to allow for creating static libraries

jockm:
Adds option to build Gravity as a static library. Also fixes minor bug in the `clean` target that didn't use the platform specific shared library name

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,8 @@ GRAVITY_SRC = src/cli/gravity.c
 EXAMPLE_SRC = examples/example.c
 
 CC ?= gcc
+AR ?= ar
+RANLIB ?= ranlib
 SRC = $(wildcard $(COMPILER_DIR)*.c) \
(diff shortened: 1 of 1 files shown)


marcobambini:
Applied in a8ea371, adapted to the Makefile as it stands today.

The target is in, with three changes to your version:

* it hangs off `$(OBJ)` rather than `gravity`, matching what `lib` does now, so the archive does not drag in the CLI's `main`
* `ar rcs` on its own — `s` already does what the separate `ranlib` call was for
* the commented-out `$(CC) -shared` line is dropped

`AR` is a built-in make variable, so `make staticlib AR=<triple>-ar` works for cross builds without declaring it in the file.

The `clean` bug you mention is fixed too, just not by this patch: `clean` now removes the shared library under each of its platform-specific names, plus `libgravity.a`.

Documented in README.md and CLAUDE.md. Verified by linking `examples/example.c` against the archive. Thanks!

PR #429, opened 2024-09-04
fixing DEFAULT_OUTPUT to one that is in --help

nerfur:
gravity.g -> gravity.json as in --help

### DIFF
--- a/src/cli/gravity.c
+++ b/src/cli/gravity.c
@@ -13,7 +13,7 @@
 #include "gravity_vm.h"
 #include "gravity_opt_env.h"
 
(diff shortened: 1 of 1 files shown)


marcobambini:
Thanks — the two really were out of sync.

It got fixed the other way round in 1236a3c: `gravity.g` is the name that was kept, and the usage text now prints `DEFAULT_OUTPUT` itself so the two cannot drift again. README.md and CLAUDE.md were saying `gravity.json` as well and were corrected at the same time.

`gravity.g` is what the compiled bytecode is called everywhere else, even though the container format happens to be JSON, so the help text was the side that was wrong.

Closing as fixed.

PR #439, opened 2026-05-25
Change CMake install instructions to be more readily packagable.

eigenwhat:
Hi! I was looking to try out gravity for a project and found that the install instructions (as in CMake `install()` instructions) were missing a few things for it to be packaged as a Conan package. I happen to use Conan, but I imagine this would also help `vcpkg` users as well. While I could just hot patch the build scripts at package time (or have it copy files manually), it's usually better for everyone if the package recipe can rely on the library's installation instructions instead so they don't fall out of sync. Additionally, I saw an issue or two open regarding how the install step doesn't copy headers, and that is one of the things I address in this pull request.

Short summary:

- Install now copies headers.
- Build considers `BUILD_SHARED_LIBS` when deciding to build `gravityapi` target.
- Install now works when CLI is disabled.
- Local in-source install on Windows is now a CMake option.
- Install paths propagated to parent scope so CLI install can see them.
- Install generates CMake find modules.

I tried to leave the current ways of using the library intact while allowing a package manager to switch to a more "orthodox" CMake install. In particular, the choice to have CMake install to a local directory on Windows is now gated behind an option that is ON by default. I also conditionally build the shared `gravityapi` target based on `BUILD_SHARED_LIBS`, which is a CMake variable usually used to swap between static/dynamic builds. Since this project builds both kinds as separate targets, I use it to decide whether `gravityapi` should be built. (`gravityapi_s` is built unconditionally.) This is defaulted to ON to match old behavior when not overridden explicitly.

A new addition is the automatic generation of find modules. It defines a `gravity-config.cmake` file which provides `gravityapi`, `gravityapi_s`, and/or `gravity` depending on build options. This now means that if installed on the *system*, gravity can now be used by running:

```cmake
find_package(gravity CONFIG REQUIRED)
target_link_libraries(my_target PUBLIC gravity::gravityapi) # or gravity::gravityapi_s
```

This would also be the manner in which a user would include it in their CMake project using a package manager.


marcobambini:
Thanks @eigenwhat!

PR #452, opened 2026-08-05
fix(json): avoid signed overflow on integers too large for int64

### LINKED ISSUE #447: [Bug]signed 64-bit integer overflow in json_parse_ex when decoding long digit runs (CWE-190)
## Summary

The vendored udp/json-parser decodes integer tokens digit-by-digit with the accumulator `top->u.integer = (top->u.integer * 10) + (b - '0')` (`src/utils/gravity_json.c:1126`). Any JSON input containing an integer literal with more than 19 significant decimal digits (e.g. twenty `2`s or twenty `9`s) makes this **signed 64-bit multiplication overflow**, which is undefined behavior. In a build with `-fsanitize=undefined -fno-sanitize-recover=all`, UBSan detects `integer multiplication overflowed` and **traps with SIGILL**.

- **Affected versions**: Gravity v0.9.7 / HEAD `17a280b83c10424597bf4d130d208b72de731c2a`
- **Severity**: Medium (wrong value / data corruption; SIGILL DoS in UBSan builds)
- **CWE**: CWE-190 (Integer Overflow)

### Affected code

```c
/* src/utils/gravity_json.c:1126 */
top->u.integer = (top->u.integer * 10) + (b - '0');   /* <-- signed 64-bit overflow on long digit runs */
```

The exponent accumulator on line 1122 (`num_e`) has the same unchecked pattern.

**Root cause**: each digit multiplies the running value by 10. Beyond 19 significant digits the value exceeds `INT64_MAX`, and signed overflow is undefined behavior in C. The build traps on it (SIGILL), and even in a non-trapping build the parser would store a wrong (wrapped) value — data corruption.

**Reachability**: 
- Directly through the public `json_parse()` API.
- Indirectly through `gravity_vm_loadbuffer()` — deserializing attacker-controlled compiled bytecode calls `json_parse()` internally.

Call chain:

```
json_parse_ex (gravity_json.c:1126, integer accumulator)
  ├─ json_parse ()                       [direct, untrusted JSON]
  └─ gravity_vm_loadbuffer () → json_parse ()   [bytecode deserialization]
```

## POC (tested on the unmodified source)

Minimal reproducer — **38 bytes** all `2`s:

```bash
echo -n '3232323232323232323232323232323232323232323232323232323232323232323232323232' | xxd -r -p > poc_rc2.json
wc -c poc_rc2.json   # -> 38
./build/gravity_fuzzer poc_rc2.json
```

## Trigger result

```
Running LLVMFuzzerInitialize ...
continue...
Illegal instruction (core dumped)    # UBSan 'integer multiplication overflowed' trap
EXIT=132 (SIGILL)
```

The trap-mode build aborts immediately; under a recoverable UBSAN build it reports `integer multiplication overflowed` at `json_parse_ex`. A 463-byte variant starting with 22 `9`s reproduces the same trap through the `gravity_vm_loadbuffer()` bytecode path.

- Replay exit code: `132` (SIGILL, UBSan trap)
- Deterministic: yes — 3 crash files (id:000004, id:000005, prev_sig04) all reproduce; 3-byte `x=0`-style minimal is 38 bytes of `2`

## Suggested fix

Guard the accumulator before each multiply, or reject digit runs longer than 19 significant digits:

```c
/* before the multiply */
if (top->u.integer > (JSON_INT_MAX - (b - '0')) / 10) goto overflow_error;
top->u.integer = (top->u.integer * 10) + (b - '0');
```

Apply the same guard to the exponent accumulator `num_e` on line 1122. Alternatively parse into a wider unsigned type with an explicit range check, or reject any integer literal exceeding 19 digits.

Socialpranker:
_Disclosure: this patch was prepared with AI assistance (Claude Opus 5). I built the project, reproduced the overflow under UBSan and checked the result against upstream json-parser myself before opening this._

Fixes #447.

### The bug

`json_parse_ex` accumulates integer literals with no bound check:

```c
top->u.integer = (top->u.integer * 10) + (b - '0');
```

A digit run longer than the `int64_t` range overflows. Signed overflow is undefined behaviour, and UBSan reports it directly:

```
src/utils/gravity_json.c:1126:55: runtime error: signed integer overflow:
2222222222222222222 * 10 cannot be represented in type 'int64_t'
```

Reachable from `JSON.parse` in Gravity code, and from `json_parse()` / `gravity_vm_loadbuffer()` in the C API.

### The fix

This file is vendored from [json-parser](https://github.com/udp/json-parser) (per the header comment), and **upstream already fixed this** — Gravity just never picked it up. So this is a backport rather than a new invention.

Upstream detects the overflow before it happens and degrades the value to a double, which is the same thing the parser already does when it sees a `.` or an exponent. Rejecting the input was not an option: such literals are valid per RFC 8259, and both Python and JavaScript accept them.

Carrying that fix over needed one more piece. Gravity told the integer and fraction parts apart by `top->type` alone, so once the value flipped to `json_double` the remaining digits were misread as a fraction and the magnitude was silently lost — `2.2e18` instead of `2.2e24`. Upstream distinguishes the two with a `flag_num_got_decimal` flag, so that came across too. Making `num_fraction` a `double` also matches upstream and removes a second overflow in the same accumulator.

Trading undefined behaviour for a quietly wrong number did not seem like a good deal, hence the slightly wider diff.

### Verification

Built with `-fsanitize=undefined -fno-sanitize-recover=signed-integer-overflow`:

| | before | after |
|---|---|---|
| `signed integer overflow` at json.c:1126 | reported | gone |

Checked against upstream json-parser compiled separately, same input — both now yield `2.22222e+24`.

Values behave as expected:

```
{"v":2222222222222222222222222}   -> 2.22222e+24
{"v":92233720368547758070}        -> 9.22337e+19
{"v":-2222222222222222222222222}  -> -2.22222e+24
{"v":9223372036854775807}         -> 9223372036854775807  (exact, still an integer)
{"v":42}                          -> 42
```

- `test/unittest/bugfix_json_integer_overflow.gravity` added, per the testing rule in CONTRIBUTING.md. It fails without the code change and passes with it — verified by reverting only the code and keeping the test.
- Full suite: **351/351, 0 failures**, on both the debug and the plain `make` build.
- Two other UBSan reports show up on this input (`gravity_json.c` null-pointer arithmetic in the two-pass sizing scheme, and a misaligned load in `gravity_hash.c`). Both reproduce on a clean `master` without this patch, so they are pre-existing and left alone.

Added myself to CONTRIBUTORS as the contributing guide asks.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -8,3 +8,4 @@ Brandon Ray   <brandon@bombsightgames.com>
 Filippo Costa <filippocosta.italy@gmail.com>
 Matan Silver  <matansilver@gmail.com>
 Martin Miralles-Cordal <axalon@rutgers.edu>
+Ivan Teresenko <0630039863abc@gmail.com>
--- a/src/utils/gravity_json.c
+++ b/src/utils/gravity_json.c
@@ -423,6 +423,11 @@ static unsigned char hex_value (json_char c)
    }
 }
 
+static int would_overflow (json_int_t value, json_char b)
+{
+   return ((JSON_INT_MAX - (b - '0')) / 10 ) < value;
+}
+
 typedef struct
 {
    unsigned long used_memory;
@@ -583,7 +588,8 @@ static const long
    flag_num_e_got_sign   = 1 << 11,
    flag_num_e_negative   = 1 << 12,
    flag_line_comment     = 1 << 13,
-   flag_block_comment    = 1 << 14;
+   flag_block_comment    = 1 << 14,
+   flag_num_got_decimal  = 1 << 15;
 
 json_value * json_parse_ex (json_settings * settings,
                             const json_char * json,
@@ -596,7 +602,7 @@ json_value * json_parse_ex (json_settings * settings,
    json_state state = EMPTY_STATE_STRUCT;
    long flags;
    long num_digits = 0, num_e = 0;
-   json_int_t num_fraction = 0;
+   double num_fraction = 0;
 
    /* Skip UTF-8 BOM
     */
@@ -1123,11 +1129,25 @@ json_value * json_parse_ex (json_settings * settings,
                         continue;
                      }
 
+                     if (would_overflow (top->u.integer, b))
+                     {
+                        json_int_t integer = top->u.integer;
+                        -- num_digits;
+                        -- state.ptr;
+                        top->type = json_double;
+                        top->u.dbl = (double)integer;
+                        continue;
+                     }
+
                      top->u.integer = (top->u.integer * 10) + (b - '0');
                      continue;
                   }
 
-                  num_fraction = (num_fraction * 10) + (b - '0');
+                  if (flags & flag_num_got_decimal)
+                     num_fraction = (num_fraction * 10) + (b - '0');
+                  else
+                     top->u.dbl = (top->u.dbl * 10) + (b - '0');
+
                   continue;
                }
 
@@ -1153,6 +1173,7 @@ json_value * json_parse_ex (json_settings * settings,
                   top->type = json_double;
                   top->u.dbl = (double) top->u.integer;
 
+                  flags |= flag_num_got_decimal;
                   num_digits = 0;
                   continue;
                }
@@ -1166,7 +1187,7 @@ json_value * json_parse_ex (json_settings * settings,
                         goto e_failed;
                      }
 
-                     top->u.dbl += ((double) num_fraction) / (pow (10.0, (double) num_digits));
+                     top->u.dbl += num_fraction / (pow (10.0, (double) num_digits));
                   }
 
                   if (b == 'e' || b == 'E')
--- a/src/utils/gravity_json.h
+++ b/src/utils/gravity_json.h
@@ -96,6 +96,10 @@ void        json_set_option (json_t *json, json_opt_mask option_value);
    #endif
 #endif
 
+#ifndef JSON_INT_MAX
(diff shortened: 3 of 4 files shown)


marcobambini:
Thanks for the careful write-up, and for checking the result against upstream rather than just silencing the sanitizer.

The undefined behaviour is already gone from master. #447 was fixed a few hours before this landed, in 9b337c3, which added the same bound upstream uses — right where your diff adds `would_overflow`:

```c
if (top->u.integer > (JSON_INT_MAX - (b - '0')) / 10)
{  snprintf (error, sizeof(error), "%d:%d: Integer literal is out of range", line_and_col);
   goto e_failed;
}
```

`JSON_INT_MAX` is already in `gravity_json.h`, and `num_fraction` got the same treatment (it drops the digits that fall past the precision of a double and keeps `num_digits` in step). This branch is based on e1f38c2, which predates that commit, so the patch no longer applies. UBSan is clean on `{"v":2222222222222222222222222}` on master today.

What is left is not the overflow but a policy difference: master rejects an out-of-range literal and fails the parse (`JSON.parse` returns `null`), while your patch degrades the value to a double the way upstream json-parser, JavaScript and Python do. The RFC 8259 argument for that is fair, and it is worth revisiting if it ever bites real interop — but it is a semantic change to a path that was just deliberately tightened, not a fix for the UB, so it should be its own discussion rather than ride along with a CWE-190 fix.

Closing as already fixed. Thanks for the report and for the reproduction steps.

PR #456, opened 2026-09-10
Fix ternary operators emitting incomplete instructions

### LINKED ISSUE #455: Ternaries not working as expected
I'm having some very strange results when evaluating ternaries.  It seems that local variables are not correctly read/converted when passed as part of the branch expressions unless they are part of another expression (or the result of a function call or cast).  Variables from any other scope (upvalues, globals, etc.) work fine.

```swift
var foo:Int = 20
System.print("Expecting: 20, 1, 1, 20, 21, 1, 1, 21")
System.print(true ? foo : 1)
System.print(true ? 1 : foo)
System.print(false ? foo : 1)
System.print(false ? 1 : foo)
System.print(true ? foo + 1 : 1)
System.print(true ? 1 : foo + 1)
System.print(false ? foo + 1 : 1)
System.print(false ? 1 : foo + 1)
```

I'm getting:
```
Expecting: 20, 1, 1, 20, 21, 1, 1, 21
true
20
1
20
21
1
1
21
```

swoolcock:
When the sole contents of the stmt or elsestmt parts of a ternary are a local identifier, no instruction is emitted to assign it to the register used by cond.  This means that cond is considered to be the result of the ternary.

This fix ensures that when pushing the local identifier onto the stack, it also emits a MOVE operation.

The ternary operator unit test has been updated to cover this case.

Resolves #455 

---

Note that I am unsure if this is the best or correct way to solve this issue, however all the existing unit tests pass fine locally and with the same output.  The updated ternary operator unit test fails without this PR with: `RUNTIME ERROR: Right operand must be a number (use the number() method).`


marcobambini (changes requested):
Thanks for tracking this down and for the clear write-up. The bug is real and your test reproduces it well (on master, `true ? a : b` with two locals returns `b`).

The fix is in the wrong place, though. Changing `visit_identifier_expr` means **every** read of a local now emits `MOVE temp, local`, not just reads inside a ternary. The optimizer removes some of these MOVEs but not all. For example, redundant MOVEs remain on call arguments (`g(a, s)`) and on subscripts (`list[a]`). Across the 353 unit tests, total bytecode grows by about 9.4% and `ntemp` goes up in many functions. That's a significant cost for a problem that only affects the ternary.

The root cause is in `visit_flow_ternary_stmt`. It assumes both branches leave their result in the same temp register, and that fails when a branch is a local, `self` or `super` (`super` is pushed directly in `visit_keyword_expr` and would still be broken with this PR). A more targeted fix is to reserve the result register up front and have each branch move its value into it only when needed:

```c
    visit(node->cond);
    reg = ircode_register_pop(code);
    if (reg == REGISTER_ERROR) report_error(self, (gnode_t *)node, "Invalid ternary condition expression.");
    ircode_add(code, JUMPF, reg, label_false, 0, LINE_NUMBER(node));

    // allocate the result register up-front so that both branches store into it
    // (a branch can leave its value in a non-temp register, like a local variable)
    uint32_t dest = ircode_register_push_temp(code);

    visit(node->stmt);
    reg = ircode_register_pop(code);
    if (reg == REGISTER_ERROR) report_error(self, (gnode_t *)node, "Invalid ternary left stmt expression.");
    if (reg != dest) ircode_add(code, MOVE, dest, reg, 0, LINE_NUMBER(node));
    ircode_add(code, JUMP, label_final, 0, 0, LINE_NUMBER(node));

    ircode_marklabel(code, label_false, LINE_NUMBER(node));
    visit(node->elsestmt);
    reg = ircode_register_pop(code);
    if (reg == REGISTER_ERROR) report_error(self, (gnode_t *)node, "Invalid ternary right stmt expression.");
    if (reg != dest) ircode_add(code, MOVE, dest, reg, 0, LINE_NUMBER(node));
    ircode_marklabel(code, label_final, LINE_NUMBER(node));
```

With this change and your updated `ternary_operator.gravity`:
- all unit tests and loadbuffer tests pass
- bytecode size is essentially unchanged (+16 hex chars across the whole suite, only where ternaries are used)
- nested ternaries, `self`/`super` in a branch, and a local reassigned after the ternary all behave correctly

It also removes the reliance on both branches happening to get the same register, so the long comment at the top of the function can be dropped.

Could you swap the `visit_identifier_expr` change for this one? Keep your test and the CONTRIBUTORS entry as they are. It would also be worth adding a `super` case to the test. Thanks!

swoolcock:
Will do when I get access to the internet on my laptop.

swoolcock:
I've made the requested changes but I haven't added super/self to the test case as I wasn't sure the best way to do it.

I tried something along the lines of:
```swift
class Foo {
    func foo() {
        return "S";
    }
}
class Bar : Foo {
    func foo() {
        return "s";
    }
}
var result = (true ? super : self).foo();
```
But it was always resolving to `self`, since you can't really put a reference to `super` on the stack, it's just magic to avoid a callvirt.

Hopefully this change is enough.
My endgoal is to add Gravity support to TIC-80, and this ternary bug is the only blocker I had.

Edit: I forgot to remove the comment, I'll do that tonight.

